PR #47, opened 2020-09-23
chore: build ic-ref with nix

nomeata:
Hans closed it in https://github.com/dfinity-lab/agent-rust/pull/38 but I want to at least get it to work to know that we can do it this way, if need be. (E.g. if buliding the BLS C library and its bindings in `ic-ref` becomes too tricky for plain cabal.)


nomeata:
Looks like I can no longer push to my branch

hansl:
Are you using a cache for Nix? 1h to build but I expect this will ONLY be the first time, right?

.github/workflows/ic-ref.yml
```diff
@@ -27,56 +28,31 @@ jobs:
...
 
-      - name: Cache ~/.cabal/store
-        uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cabal/store
-          key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-store
-
-      - uses: actions/setup-haskell@v1.1
+      - name: Cache /nix/store
```
hansl, line 31: Does nix run garbage collection once in a while?

nomeata:
> Are you using a cache for Nix? 1h to build but I expect this will ONLY be the first time, right?

I hope so, but Github Actions don’t write to the cache from failed builds (and I couldn't find out how to change that), so it needs to succeed once before we see if the cache works.

In my own projects I use https://cachix.org/ for caching; that creates a per-project nix cache that even my contributors can use, but that is not a viable option here as long as we pull in internals.

nomeata:
Look, it’s green! So it’s possible at least :-)

hansl (approved):
I'm okay with this, only problem I can see is nix store cache filling up (with older versions of ic-ref) and never being garbage collected. Do we know if there's a way to manage that?

nomeata:
Hmm, there might be a way to `nix-store --gc` at the end in a way that keeps all build dependencies alive, but I’d have to dig deeper.

We don’t have to merge this now as long as the other way works fine.

PR #51, opened 2020-09-24
feat: add set controller to the management canister

p-shahi:


ref-tests/src/utils.rs
```diff
@@ -25,7 +25,7 @@ pub async fn create_identity() -> Result<BasicIdentity, String> {
     ))
 }
 
-pub async fn create_agent() -> Result<Agent, String> {
+pub async fn create_agent(identity: BasicIdentity) -> Result<Agent, String> {
```
p-shahi, line 28: Changed this here, now requring `create_identity()` be called separately instead of implicitly in `create_agent()`
This is to allow getting the Principal for that identity (which otherwise would have been private.)
Open to suggestion if this should be done a different way

ref-tests/tests/ic-ref.rs
```diff
@@ -130,7 +131,9 @@ mod management_canister {
                 .await?;
 
             // Each agent has their own identity.
-            let other_agent = create_agent().await?;
+            let other_agent_identity = create_identity().await?;
+            let other_agent_principal = other_agent_identity.sender()?;
```
p-shahi, line 135: see: https://github.com/dfinity/agent-rs/pull/51/files#r494050809

hansl (approved):

p-shahi:
Hm seems merging is still blocked @hansl

PR #54, opened 2020-09-25
fix: use candid Natural instead of fixed bytes numbers in install

hansl:
And added a test for memory_allocation and compute_allocation.


ref-tests/tests/ic-ref.rs
```diff
@@ -455,3 +455,57 @@ mod simple_calls {
         })
     }
 }
+
+mod extras {
```
p-shahi, line 459: 👍

p-shahi (approved):

PR #56, opened 2020-10-06
feat: add query call builder, call_raw on sync calls and wallet canister interface

hansl:


ic-utils/src/call.rs
```diff
@@ -132,37 +147,42 @@ where
...
-    where
-        R: for<'de> ArgumentDecoder<'de> + Send + Sync,
-    {
-        self.agent
-            .query(&self.canister_id, &self.method_name)
-            .with_arg(&self.method_name)
-            .call()
-            .await
-            .and_then(|r| decode_args(&r).map_err(|e| AgentError::CandidError(Box::new(e))))
+    async unsafe fn call_raw(self) -> Result<Vec<u8>, AgentError> {
```
ericswanson-dfinity, line 167: Is the whole reason for `unsafe` here to discourage direct calls to call_raw?

It seems like as a side-effect, this function could accumulate unsafe code.

Could we change this function to a one-liner that forwards to a function that is not marked `unsafe` ?

ericswanson-dfinity: Or is there some other way to hide `call_raw` ?

ic-utils/src/interfaces/wallet.rs
```diff
@@ -0,0 +1,184 @@
...
+    ) -> impl 'agent + AsyncCall<()> {
+        self.update_("authorize").with_arg(custodian).build()
+    }
+
+    /// Deauthorize a custodian.
+    pub fn deauthorize<'canister: 'agent>(
+        &'canister self,
+        custodian: Principal,
+    ) -> impl 'agent + AsyncCall<()> {
+        self.update_("authorize").with_arg(custodian).build()
```
ericswanson-dfinity, line 144: should this be the following?
```suggestion
        self.update_("deauthorize").with_arg(custodian).build()
```

ic-utils/src/call.rs
hansl: Feature flag would be another way. You have to "opt in" at the crate level, which is fine I guess.

hansl: Done. PTAL

ericswanson-dfinity (approved):

PR #60, opened 2020-10-09
feat: call forwarding from a wallet canister

hansl:
This can happen either by passing raw bytes arguments (using the previous `call()`), or by building a Call and passing the Call in.

### DIFF
--- a/.github/workflows/ic-ref.yml
+++ b/.github/workflows/ic-ref.yml
@@ -30,6 +30,11 @@ jobs:
           token: ${{ secrets.IC_REF_TOKEN }}
           path: ic-ref
           ref: ${{ matrix.spec }}
+      - uses: actions/checkout@v2
+        with:
+          repository: 'dfinity/wallet-rs'
+          path: wallet-rs
+          ref: latest
 
       - name: Cache ~/.cabal/store
         uses: actions/cache@v2
@@ -74,13 +79,20 @@ jobs:
           cargo build --target wasm32-unknown-unknown --release
           cp target/wasm32-unknown-unknown/release/universal_canister.wasm $HOME/canister.wasm
 
-      - name: Run Tests
+      - name: Build wallet canister
+        run: |
+          cd wallet-rs/wallet
+          sh build.sh
+          cp wallet/target/wasm32-unknown-unknown/release/wallet-opt.wasm $HOME/wallet.wasm
+
+      - name: Run Integration Tests
         run: |
           set -ex
           $HOME/bin/ic-ref --pick-port --write-port-to $HOME/ic_ref_port &
           sleep 1
           export IC_REF_PORT=$(cat $HOME/ic_ref_port)
           export IC_UNIVERSAL_CANISTER_PATH=$HOME/canister.wasm
+          export IC_WALLET_CANISTER_PATH=$HOME/wallet.wasm
           cd main/ref-tests
           cargo test --all-features -- --ignored
           killall ic-ref
--- a/ic-utils/src/call.rs
+++ b/ic-utils/src/call.rs
@@ -207,6 +207,8 @@ impl<'agent, Out> AsyncCaller<'agent, Out>
 where
     Out: for<'de> ArgumentDecoder<'de> + Send + Sync,
 {
+    /// Build an UpdateBuilder call that can be used directly with the [Agent]. This is
+    /// essentially downleveling this type into the lower level [ic-agent] abstraction.
     pub fn build_call(self) -> Result<UpdateBuilder<'agent>, AgentError> {
         let mut builder = self.agent.update(&self.canister_id, &self.method_name);
         self.expiry.apply_to_update(&mut builder);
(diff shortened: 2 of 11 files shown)


ref-tests/Cargo.toml
```diff
@@ -6,12 +6,13 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+candid = "0.6.7"
 delay = "0.3.0"
 ic-agent = { path = "../ic-agent" }
 ic-utils = { path = "../ic-utils", features = ["raw"] }
 ring = "0.16.11"
 tokio = "0.2.22"
+serde = { version = "1.0.101", features = ["derive"] }
```
ericswanson-dfinity, line 15: alphabetical order?

```suggestion
serde = { version = "1.0.101", features = ["derive"] }
tokio = "0.2.22"
```

ref-tests/tests/integration.rs
```diff
@@ -75,3 +79,53 @@ fn canister_query() {
         Ok(())
     })
 }
+
+#[ignore]
+#[test]
```
ericswanson-dfinity, line 84: I'd consider this adjacent to this PR, since there are already two other tests marked `#[ignore]`, but I notice that I am wondering what process to follow if I wanted to run these, since presumably `cargo test` is not it.

.github/workflows/ic-ref.yml
```diff
@@ -74,13 +79,20 @@ jobs:
           cargo build --target wasm32-unknown-unknown --release
           cp target/wasm32-unknown-unknown/release/universal_canister.wasm $HOME/canister.wasm
 
-      - name: Run Tests
+      - name: Build wallet canister
+        run: |
+          cd wallet-rs/wallet
+          sh build.sh
```
ericswanson-dfinity, line 85: where is `build.sh` ?

ref-tests/tests/integration.rs
ericswanson-dfinity: Nevermind, I see the github workflow now.

ic-utils/src/interfaces/wallet.rs
```diff
@@ -163,22 +168,53 @@ impl<'agent> Canister<'agent, Wallet> {
 
     /// Forward a call to another canister, including an amount of cycles
     /// from the wallet.
-    pub fn call<'canister: 'agent, Out>(
+    pub fn call<'canister: 'agent, Out, M: ToString>(
```
ericswanson-dfinity, line 171: I'm curious: is this `M: ToString` here instead of `&str` because it's sometimes annoying to have to call `.to_string()/.as_str()/.whatever()`?

ref-tests/tests/integration.rs
hansl: I explain it somewhere (but can definitely make it clearer) that ignored tests are run against the ic-ref. Because they require a replica, they are ignored and ran separately. I'll make it clearer in docs.

.github/workflows/ic-ref.yml
hansl: in the wallet repo

ic-utils/src/call.rs
```diff
@@ -199,7 +207,7 @@ impl<'agent, Out> AsyncCaller<'agent, Out>
 where
     Out: for<'de> ArgumentDecoder<'de> + Send + Sync,
 {
-    fn build_call(self) -> Result<UpdateBuilder<'agent>, AgentError> {
+    pub fn build_call(self) -> Result<UpdateBuilder<'agent>, AgentError> {
```
ericswanson-dfinity, line 210: Now that this is `pub`, should it have one of those `///` comments?

ericswanson-dfinity (approved):
LGTM with a couple truly minor suggestions.

Fair warning that I'm finding the lifetimes to be impenetrable, especially when there are multiple involved.  Another way of saying that: with respect to lifetimes given my understanding of them, I consider my approval to hold no more weight than a rubber-stamp.

ic-utils/src/interfaces/wallet.rs
hansl: Yeah this accepts any type which have the trait `ToString`, which includes `&str` and `String` (and maybe others?). So this is a shortcut to say "we accept both". I need to go through the whole repo and change our strings to that. I'm going to do a clean of these before version 0.2, which should also include an `IntoPrincipal` that can be either String, &str, or another Principal (or a `Canister<...>` for example).

hansl: Also including things like a typed `RequestId` and better ways to wait on Requests (moving `call_and_wait()` into `call().and_wait()` (wait being on RequestId).

As I'm working between repos I see a lot more cracks like these that can improve the API greatly.

ic-utils/src/call.rs
hansl: Done. Thanks!

ref-tests/Cargo.toml
hansl: Done separately with the rest of the changes.

ref-tests/tests/integration.rs
hansl: Oh it's at the top of this file in a comment.

hansl:
@ericswanson-dfinity Not ignoring your comment; lifetimes were added in this PR because there was some inconsistencies in them, and we will keep adding them as we find them. I don't think it's likely to fix all lifetimes properly since broader lifetimes won't error out until you try to restrict them.

PR #65, opened 2020-10-14
feat: DER-encode public keys

ericswanson-dfinity:
Wrap Ed25519 keys in DER encoding.  Calculate principals based on the DER-encoding (note that this changes principal ids).

Implements https://github.com/dfinity/sdk/issues/1126

Prerequisites:
- [ic-ref #200](https://github.com/dfinity-lab/ic-ref/pull/200) to be merged.
- sdk updated to at least replica https://github.com/dfinity-lab/dfinity/commit/0e1f85863cf54ab2d501ef1c3e276dd26265aef9

See [sdk #1127](https://github.com/dfinity/sdk/pull/1127) for proof of e2e tests passing

This is marked blocked because we have decided not to merge it until shortly before the Mercury release.  This way we do not have to handle both raw Ed25519 keys and DER-wrapped keys (which will have a different principal).


ericswanson-dfinity:
See also https://github.com/dfinity/sdk/pull/1127

nomeata:
The fix would work (see #66), but it will not be picked up by `pull_request_target`, i.e. only when it reached `next`. This probably needs repo admin intervention.

ic-agent/src/identity/mod.rs
```diff
@@ -7,6 +7,7 @@ pub use basic::{BasicIdentity, PemError};
 
 #[derive(Clone, Debug)]
 pub struct Signature {
+    pub der_encoded_public_key: Vec<u8>,
```
hansl, line 10: This would have to be added to all identity signers. Is this required from the spec now?

ericswanson-dfinity: We could leave the field `public_key`, but it would have to be understood that they are DER-encoded.

ericswanson-dfinity: Removed this field, and noted that `public_key` must be DER-encoded.

hansl:
The error from ic-ref is `public key has wrong length 44, expected 32`. Is DER accepted as spec 0.11?

ericswanson-dfinity:
@hansl this PR will need https://github.com/dfinity-lab/ic-ref/pull/200 in order to work

nomeata:
You should be able to run the `ic-ref` CI against a branch, right?

ericswanson-dfinity:
@nomeata Sure, I'll try replacing this with a SHA from the branch
```
          ref: ${{ matrix.spec }}
```

hansl (approved):

PR #64, opened 2020-10-13
chore: ic-ref CI: Use ghc-8.8.4

nomeata:
to match what’s in nixpkgs-20.09

PR #66, opened 2020-10-16
chore: ic-ref CI: Use ghc-8.8.4

nomeata:
to match what’s in nixpkgs-20.09

(This is https://github.com/dfinity/agent-rs/pull/64 again, this time from within the repo, to test the workflow changes.)

### DIFF
--- a/.github/workflows/ic-ref.yml
+++ b/.github/workflows/ic-ref.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - next
   pull_request_target:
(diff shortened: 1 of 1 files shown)


nomeata:
Ok, `pull_request` passes now.  Still unsure how to merge this: The `pull_request_target` fails (and will continue to fail). And not sure if you actually want the `pull_request` trigger in `next`?

This might need an admin push to `next` or something. Will leave to @hansl .

PR #67, opened 2020-10-22
feat: use ic-ref master for response authentication support

ericswanson-dfinity:


p-shahi (approved):

p-shahi:
Looks like ic-ref expects the spec version to be `∞`

nomeata:
> Looks like ic-ref expects the spec version to be ∞

I’d say `ic-ref` “reports”, not “expects” the version.  But the `master` branch of the `ic-ref` repo is probably unsuited, as it contains an undefined state of the implementation (the `master` branch is where we collect finished _designs_ in the spec document).

The implementation there is currently identical to the one on `release-0.12`, and you _probably_ want to use either that, or maybe one of the feature branches (e.g. `joachm/impl-webauthn`, sorry for the typo).

PR #71, opened 2020-10-30
feat: Hash tree, for certificates within response auth.

matthewhammer:
Spec: https://hydra.dfinity.systems/latest/dfinity-ci-build/ic-ref.pr-218/public-spec/1/index.html#_encoding_of_certificates

### DIFF
--- a/ic-agent/src/hash_tree.rs
+++ b/ic-agent/src/hash_tree.rs
@@ -7,19 +7,19 @@
 - [Reference implementation in the replica](https://github.com/dfinity-lab/dfinity/tree/master/rs/crypto/tree_hash)
 
 */
-use serde::{export::Formatter, ser::SerializeSeq, Deserialize, Serialize, Serializer};
-use serde_bytes::Bytes;
+use serde::{export::Formatter, Deserialize, Serialize};
+
+use openssl::sha::Sha256;
 
 use std::convert::TryFrom;
 use std::fmt;
 use std::fmt::Debug;
+use std::convert::TryInto;
 
 const DOMAIN_HASHTREE_LEAF: &str = "ic-hashtree-leaf";
 const DOMAIN_HASHTREE_EMPTY_SUBTREE: &str = "ic-hashtree-empty";
 const DOMAIN_HASHTREE_NODE: &str = "ic-hashtree-labeled";
 const DOMAIN_HASHTREE_FORK: &str = "ic-hashtree-fork";
-
-use openssl::sha::Sha256;
 
 /// Type alias for a sha256 result (ie. a u256).
 type Sha256Hash = [u8; 32];
@@ -184,35 +184,36 @@ impl AsRef<[u8]> for Digest {
     }
 }
 
-pub struct Hasher(Sha256Hash);
+pub struct Hasher(Sha256);
 
 impl Hasher {
     pub fn for_domain(domain: &str) -> Self {
         assert!(domain.len() < 256);
-        let mut hasher = Self(Sha256g::new());
+        let mut hasher = Self(Sha256::new());
         hasher.update(&[domain.len() as u8][..]);
         hasher.update(domain.as_bytes());
         hasher
     }
     pub fn update(&mut self, bytes: &[u8]) {
-        self.0.write(bytes);
+        self.0.update(bytes);
     }
     pub fn finalize(self) -> Digest {
         Digest(self.0.finish())
     }
 }
 
+/// Public spec [hash-tree](https://hydra.dfinity.systems/latest/dfinity-ci-build/ic-ref.pr-218/public-spec/1/index.html#_encoding_of_certificates)
 #[derive(Clone, PartialEq, Eq, Debug)]
-pub enum MixedHashTree {
+pub enum HashTree {
     Empty,
-    Fork(Box<(MixedHashTree, MixedHashTree)>),
-    Labeled(Label, Box<MixedHashTree>),
+    Fork(Box<(HashTree, HashTree)>),
+    Labeled(Label, Box<HashTree>),
     Leaf(Vec<u8>),
     Pruned(Digest),
 }
 
-impl MixedHashTree {
-    /// Recomputes root hash of the full tree that this mixed tree was
+impl HashTree {
+    /// Recomputes root hash of the full tree that this hash tree was
     /// constructed from.
     pub fn digest(&self) -> Digest {
         match self {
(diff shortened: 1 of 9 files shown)


matthewhammer:
I will resume this PR starting tomorrow.  (cc @ericswanson-dfinity)

matthewhammer:
Question for anyone:

How do I get the best equivalent of this crate dependency here? `ic_crypto_sha256`

I'll check what the rest of the agent uses, if anything. cc @ericswanson-dfinity @chenyan-dfinity

chenyan-dfinity:
I believe it's used in requestId, so here: https://github.com/dfinity/agent-rs/blob/next/ic-agent/src/request_id/mod.rs#L130

matthewhammer:
I see, that crate (`ic-agent`) uses `openssl = "0.10.24"`

But so far, `ic-types` does not.  I could use it too, or use what's there already (for some other reason, presumably): `sha2 = "0.9.1"`

Perhaps it doesn't matter, but I'm also wondering about best practices around dependencies.  If this can be a teachable moment for me, that's great. : )

I will use the one that's there, since it's one fewer changes to the `Cargo.toml`

ericswanson-dfinity:
agent-rs already has a dependency that provides `Sha256`.  Can we use that?

chenyan-dfinity:
Certificate doesn't seem to be something external facing, why do we want to implement in `ic-types` and not `ic-agent`?

matthewhammer:
> Certificate doesn't seem to be something external facing, why do we want to implement in ic-types and not ic-agent?

Because "Certificates should not be public (yet)" (@hansl and @ericswanson-dfinity in SDK standup).

The rationale is that ic-agent is where we can put this to hide it within a crate, and avoid a public type with a cross-crate dependency.  I disagree with this rationale, but will follow it now.

ericswanson-dfinity (approved):

matthewhammer:
@hansl Please help me merge this, to unblock @ericswanson-dfinity

ericswanson-dfinity:
bors merge

matthewhammer:
Moving on to #75

ericswanson-dfinity:
@hansl this is blocking

ic-agent/src/hash_tree.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// Vec<u8> is typically 3 machine words long (pointer + size + capacity) which
+/// is 24 bytes on amd64.  It's a good practice to keep enum variants of
+/// approximately the same size. We want to optimize for labels of at most 32
+/// bytes (as we will have many labels that are SHA256 hashes).
+const SMALL_LABEL_SIZE: usize = 32;
+
+/// This type hides the implementation of Label.
+#[derive(Clone)]
+enum LabelRepr {
```
hansl, line 44: Why do we need this type? If you want to save RAM, we could just use https://docs.rs/smallvec/1.4.2/smallvec/ ?

ic-agent/src/hash_tree.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+    pub fn update(&mut self, bytes: &[u8]) {
+        self.0.update(bytes);
+    }
+    pub fn finalize(self) -> Digest {
+        Digest(self.0.finish())
+    }
+}
+
+/// Public spec [hash-tree](https://hydra.dfinity.systems/latest/dfinity-ci-build/ic-ref.pr-218/public-spec/1/index.html#_encoding_of_certificates)
```
hansl, line 202: Can we point the URL to the published (still internal) version of the spec that implements it? Something like https://docs.dfinity.systems/public/v/0.12.2/. Using a link to a rendered PR is bad form IMO (and the PR might be updated or rejected).

ic-agent/src/hash_tree.rs
```diff
@@ -0,0 +1,274 @@
+/** Hash-tree
+
+#### Resources:
+
+- [Public spec link](https://hydra.dfinity.systems/latest/dfinity-ci-build/ic-ref.pr-218/public-spec/1/index.html#_encoding_of_certificates)
```
hansl, line 5: Can we point the URL to the published (still internal) version of the spec that implements it? Something like https://docs.dfinity.systems/public/v/0.12.2/. Using a link to a rendered PR is bad form IMO (and the PR might be updated or rejected).

ic-agent/src/hash_tree.rs
nomeata: There a small bits in that PR that are not in a  published version yet, waiting for https://github.com/dfinity-lab/dfinity/pull/6611/ to merge to do a release (we have fallen a bit into the habit of synchronizing the dfinity `master` merge with the `ic-ref` release). We should be able to update this later this week, I hope.

ic-agent/src/hash_tree.rs
matthewhammer: Yep, that's fine with me.

ic-agent/src/hash_tree.rs
matthewhammer: No real gain to your suggestion, just premature optimization.

Inductive structures are easier to understand, and more closely follow the public spec.

The code here adapts somewhat more complex replica code that does the same thing.  Argue with them if you really care.  I don't think we can afford to.

ic-agent/src/lib.rs
```diff
@@ -105,3 +105,6 @@ pub mod request_id;
 pub use agent::{agent_error::AgentError, nonce::NonceFactory, Agent, PasswordManager};
 pub use identity::{Identity, Signature};
 pub use request_id::{to_request_id, RequestId, RequestIdError};
+
+// FIXME: not actually public
+pub mod hash_tree;
```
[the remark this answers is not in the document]
matthewhammer: Sure if you care.  Feel free to change it here (this PR), or later.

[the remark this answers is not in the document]
matthewhammer: *As I general rule, I want all types that are even remotely related to the public spec to be public(ly visible) to all users, even if they change later in subsequent versions.*

I will never change this stance, and we will never, ever agree.  So, we need not argue further.  Apparently you have the power and the responsibility that goes with it here, @hansl, so feel free to edit this type and others as needed.

ic-agent/src/hash_tree.rs
matthewhammer: The URL is not perfect yet.

But let's leave it here (it continues to have value for us, I think?), and update it later when we can.

It's okay if this PR is not perfect. It's okay if this PR is not perfect. It's okay if this PR is not perfect.

hansl:
Closing this and will try to delete this PR. This is an open source repo and we probably don't want this to stay online.

PR #75, opened 2020-11-04
feat: Hashtree path lookup

matthewhammer:
Extends #71 with additional code, for path lookup.

This completes item (b) from our response auth checklist.

### DIFF
--- a/ic-agent/src/agent/builder.rs
+++ b/ic-agent/src/agent/builder.rs
@@ -20,10 +20,10 @@ impl AgentBuilder {
     }
 
     /// Set the URL of the [Agent].
-    pub fn with_url<S: ToString>(self, url: S) -> Self {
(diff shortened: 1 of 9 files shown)


ericswanson-dfinity (approved):

ericswanson-dfinity:
@hansl this is blocking

matthewhammer:
I hereby give @hansl permission to change this PR and #71 any way he wants so satisfy his desiderata for the repo.  I don't need to understand or agree, and I suspect that I will not.  : )

matthewhammer:
Surprised this (still) isn't merged.

Is it not blocking @ericswanson-dfinity anymore?  (If it is still critical, are the comments above really that critical to address now, in this PR?)

ericswanson-dfinity:
I

PR #76, opened 2020-11-10
feat: add a HashTree type for the certificate state

hansl:
It understands labels and can lookup paths and return blobs of those
paths, if found.

### DIFF
--- a/ic-agent/src/hash_tree.rs
+++ b/ic-agent/src/hash_tree.rs
@@ -293,10 +293,10 @@ impl HashTreeNode {
                                     LookupLabelResult::Absent
                                 }
                             }
-                            result @ _ => result,
+                            result => result,
                         }
                     }
-                    result @ _ => result,
+                    result => result,
                 }
             }
             HashTreeNode::Pruned(_) => LookupLabelResult::Unknown,
--- a/ic-agent/src/hash_tree.rs
+++ b/ic-agent/src/hash_tree.rs
@@ -124,7 +124,7 @@ impl<'de> serde::Deserialize<'de> for HashTree {
 
             fn visit_seq<V>(self, mut seq: V) -> Result<Self::Value, V::Error>
             where
-                V: SeqAccess<'de>,
+                V: de::SeqAccess<'de>,
             {
                 let tag: u8 = seq
                     .next_element()?
--- a/ic-agent/src/hash_tree.rs
+++ b/ic-agent/src/hash_tree.rs
@@ -6,13 +6,15 @@
 //!
 //! cf https://docs.dfinity.systems/public/v/0.13.1/#_encoding_of_certificates
 use openssl::sha::Sha256;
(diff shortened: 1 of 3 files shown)


ericswanson-dfinity (approved):
Thanks!

ic-agent/src/hash_tree.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// is not necessary.
+    fn lookup_label(&self, label: &Label) -> LookupLabelResult {
+        match self {
+            // If this node is a labeled node, check for the name. This assume a
+            HashTreeNode::Labeled(l, node) => match label.cmp(l) {
+                std::cmp::Ordering::Greater => LookupLabelResult::Continue,
+                std::cmp::Ordering::Equal => LookupLabelResult::Found(node.as_ref()),
+                // If this node has a smaller label than the one we're looking for, shortcut
+                // out of this search (sorted tree), we looked too far.
+                std::cmp::Ordering::Less => LookupLabelResult::Absent,
```
nomeata, line 310: :angry: Github, just ate this long comment. Second try.

```suggestion
                std::cmp::Ordering::Less => LookupLabelResult::Error,
```
I don’t think you can claim `Absent ` here; the lookup codes doesn't yet do the necessary bookkeeping.

Consider this tree:
```
Fork( Fork(Labeled("a",…), Pruned(…)), Labeled("c", …) )
```
looking up label `b` in this tree should return `Unknown` (after all, the value _might_ be in the `Pruned` tree, can't tell).

I recommend to just return `Unknown` here for now, which is good enough to unblock this project; the algorithm can be refined in a later PR.

If you want to do it right now, here is the way to do it: While you traverse the tree in order, you pass a flag around `prev_item_was_pruned`, initially `false`. If you hit `Pruned`, you turn this to `true`, when you hit `Labeled` to `false`. When hitting a label that’s larger (or the end of the tree) than the searched for, return `Absent` if the flag was `false` and `Unknown` if it was `true`.

hansl: I fixed this whole function after debugging with your examples. They helped a lot, thanks!

hansl: For reference, it wasn't Error either. It really _is_ absent, but there is another state which is Unknown when we pass a Pruned on the way. It's just a bunch of edge cases that I fixed this morning essentially.

hansl: >  While you traverse the tree in order, you pass a flag around prev_item_was_pruned

That's almost what I did. I have another LookupLabelResult value which is Unknown. Absent gets transformed into Unknown if it meets a pruned.

hansl:
Merging this per @ericswanson-dfinity  request.

ic-agent/src/hash_tree.rs
nomeata: That works too, nice work!

PR #79, opened 2020-11-16
ci: fix setup-haskell step

hansl:
See https://github.blog/changelog/2020-11-09-github-actions-removing-set-env-and-add-path-commands-on-november-16/

The setup-haskell step, previous of versions 1.1.3, was not updated properly for the change. Just using a sliding v1.1 was apparently not enough to get the latest, so we force it.

### DIFF
--- a/.github/workflows/ic-ref.yml
+++ b/.github/workflows/ic-ref.yml
@@ -46,9 +46,10 @@ jobs:
             ~/.cabal/store
           key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-store
 
-      - uses: actions/setup-haskell@v1.1
+      - uses: actions/setup-haskell@v1.1.3
         with:
           ghc-version: ${{ matrix.ghc }}
+          cabal-version: '3.2'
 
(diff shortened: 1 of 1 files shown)


hansl:
Just for reference, the `pull_request_target` will fail, and that's normal. It doesn't take the CI changes from a PR. I should figure how to filter it out from PRs that come from inside the repo.

ericswanson-dfinity (approved):

nomeata:
> Just for reference, the `pull_request_target` will fail, and that's normal. It doesn't take the CI changes from a PR. I should figure how to filter it out from PRs that come from inside the repo.

I had the same problem in an unrelated project recently, and the best thing I found was to set
```
    if: github.repository == 'foo/bar'
```

hansl:
@nomeata doesn't seem to be understood, see https://github.com/dfinity/agent-rs/runs/1408966639?check_suite_focus=true

And I'm not 100% sure this is the behaviour we want. I'm starting to rethink our whole pull_request/pull_request_target strategy (when using forks).

PR #81, opened 2020-11-17
Use provisional_create_canister_with_cycles instead of create_canister

akhi3030:
In replica, we want to restrict create_canister to be only callable by other canisters.  Hence, we need to update the agent to use
provisional_create_canister_with_cycles which provides the same semantics as create_canister does right now.

### DIFF
--- a/ic-utils/src/interfaces/management_canister.rs
+++ b/ic-utils/src/interfaces/management_canister.rs
@@ -285,7 +285,7 @@ impl<'agent> Canister<'agent, ManagementCanister> {
             canister_id: Principal,
         }
 
-        self.update_("create_canister")
+        self.update_("provisional_create_canister_with_cycles")
             .build()
             .map(|result: (Out,)| (result.0.canister_id,))
     }


hansl:
This doesn't seem supported by the spec yet. Is there an issue we can follow?

The management canister interface (all of ic-utils actually) will also be made available in WASM shortly (as soon as I can give this and the Rust CDK some love). I'll wait for the spec update but if this is meant for being called from agents, while calling another method in canisters, we should use features or arch to distinguish between the two, and leave the interface the same. WDYT?

nomeata:
> This doesn't seem supported by the spec yet. Is there an issue we can follow?

https://github.com/dfinity-lab/ic-ref/pull/234. I expect that this will become 0.14 tomorrow or so.
 
> if this is meant for being called from agents, while calling another method in canisters

I don’t think that’s the necessarily the right axis to decide which one to use. I think the choices are:

* Use `provisional_` in local development (both as a user or, if you want, from a canister)
* Use `create_canister` from canisters on production (and also in local development if you want to test canisters that create other other canisters). This can only be used from a canister, because it requires cycles to be put on the call explicitly, and requires canisters to make a choice about how much of their cycles they want to give the new canister.
 * Tell your “wallet” canister to use `create_canister` on production as a user. This is what (I assume) `dfx` will do upon `dfx deploy` for Mercury I.

This describes what I think is the plan for Mercury I. I think @akhi3030 want’s this PR to just take one, small step in that direction, so I don’t expect all of that reflected here; i am just giving an outlook.

akhi3030:
Thanks for the nice summary @nomeata .  @hansl does that sound reasonable?

hansl:
> Thanks for the nice summary @nomeata . @hansl does that sound reasonable?

It clarifies things, yes. Let's wait for the 234 PR to be merged and released, and make sure the ic-ref CI tests pass.

akhi3030:
> > Thanks for the nice summary @nomeata . @hansl does that sound reasonable?
> 
> It clarifies things, yes. Let's wait for the 234 PR to be merged and released, and make sure the ic-ref CI tests pass.

Definitely!  I just wanted to see how hard it would be to update this repo while I had a few free cycles.  Turns out, not very hard at all.

PR #83, opened 2020-11-18
feat: Updates ic-ref to 0.14

akhi3030:
Updates ic-ref to 0.14.0

### DIFF
--- a/ref-tests/src/utils.rs
+++ b/ref-tests/src/utils.rs
@@ -114,12 +114,12 @@ pub async fn create_wallet_canister(agent: &Agent) -> Result<Principal, Box<dyn 
         canister_id: Principal,
     }
 
+    // Specifying None for num_cycles will cause the canister to be created with
+    // sufficiently large number of cycles that should allow it to exist without
+    // needing to be refilled for a couple of months.
     let (Output { canister_id },) = ic00
-        .update_("dev_create_canister_with_funds")
-        .with_arg(Input {
-            num_cycles: candid::Nat::from(1_000_000_000_000u64),
-            num_icpt: candid::Nat::from(1_000u64),
-        })
+        .update_("provisional_create_canister_with_cycles")
+        .with_arg(Input { num_cycles: None })
         .build()
         .call_and_wait(create_waiter())
         .await?;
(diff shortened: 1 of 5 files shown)


ref-tests/tests/ic-ref.rs
```diff
@@ -13,7 +13,7 @@
 use ref_tests::universal_canister;
 use ref_tests::with_agent;
 
-const EXPECTED_IC_API_VERSION: &str = "0.11.1";
+const EXPECTED_IC_API_VERSION: &str = "0.14";
```
nomeata, line 16:
```suggestion
const EXPECTED_IC_API_VERSION: &str = "0.14.0";
```

akhi3030:
@hansl : could you look at this PR for a moment please?  The tests that are failing seem to be passing just fine for me when I run them locally.  I am not sure what is going wrong.

ref-tests/src/utils.rs
```diff
@@ -105,21 +105,20 @@ pub async fn create_wallet_canister(agent: &Agent) -> Result<Principal, Box<dyn
 
     #[derive(CandidType)]
     struct Input {
-        num_cycles: candid::Nat,
-        num_icpt: candid::Nat,
+        num_cycles: Option<candid::Nat>,
```
dsarlis, line 108: @akhi3030 I think you need to name this `amount` instead of `num_cycles`. `provisional_create_canister_with_cycles` expects an `amount` field based on its Candid interface:
```
provisional_create_canister_with_cycles :
    (record { amount: opt nat }) -> (record {canister_id : canister_id});
```
I'm not sure how the tests work for you locally -- I'd say they shouldn't.

akhi3030: Good catch @dsarlis.  Now updated.

dsarlis: Well, there are other failures now but at least this one was fixed.

akhi3030: Yeah, this failure is related to `cdk-rs` repo I believe.

dsarlis: Hmm, ok, from the looks of it this PR has to wait until both `cdk-rs` and `wallet-rs` are updated to use the new system api. Some tests here are using the wallet wasm and it's still importing the old api.

hansl (approved):

PR #86, opened 2020-11-19
feat: response authentication (certificate verification, root key lookup, delegations)

ericswanson-dfinity:
Implements from doc:
- Checking BLS signature
- Fetching root key from /api/v1/status for dev instances
- Hard-coding Sodium key (**still need the actual key**).  (Note doc says this would be the Mercury key)
- Support for delegations 

See also companion PR [sdk #1197](https://github.com/dfinity/sdk/pull/1197)

### DIFF
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -78,6 +78,15 @@ pub enum AgentError {
 
     #[error("The request status ({1}) at path {0:?} is invalid.")]
     InvalidRequestStatus(Vec<Label>, String),
+
+    #[error("Certificate verification failed.")]
+    CertificateVerificationFailed(),
+
+    #[error(r#"BLS DER-encoded public key must be ${expected} bytes long, but is {actual} bytes long."#)]
+    DerKeyLengthMismatch { expected: usize, actual: usize },
+
+    #[error("BLS DER-encoded public key is invalid. Expected the following prefix: ${expected:?}, but got ${actual:?}")]
+    DerPrefixMismatch{ expected: Vec<u8>, actual: Vec<u8> },
 }
 
 impl PartialEq for AgentError {
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -1,3 +1,4 @@
+use crate::agent::status::Status;
 use crate::hash_tree::Label;
 use crate::RequestIdError;
 use leb128::read;
@@ -90,8 +91,11 @@ pub enum AgentError {
     #[error("BLS DER-encoded public key is invalid. Expected the following prefix: ${expected:?}, but got ${actual:?}")]
     DerPrefixMismatch { expected: Vec<u8>, actual: Vec<u8> },
 
-    #[error("The status response did not contain a root key")]
-    NoRootKeyInStatus(),
+    #[error("The status response did not contain a root key.  Status: {0}")]
+    NoRootKeyInStatus(Status),
+
+    #[error("Could not read the root key")]
+    CouldNotReadRootKey(),
 
     #[error("Failed to initialize the BLS library")]
     BlsInitializationFailure(),
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -92,6 +92,9 @@ pub enum AgentError {
 
     #[error("The status response did not contain a root key")]
     NoRootKeyInStatus(),
+
+    #[error("Failed to initialize the BLS library")]
+    BlsInitializationFailure(),
 }
 
 impl PartialEq for AgentError {
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -82,11 +82,13 @@ pub enum AgentError {
     #[error("Certificate verification failed.")]
     CertificateVerificationFailed(),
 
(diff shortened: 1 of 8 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -28,11 +28,22 @@ use reqwest::Method;
...
 
-const DOMAIN_SEPARATOR: &[u8; 11] = b"\x0Aic-request";
+const IC_REQUEST_DOMAIN_SEPARATOR: &[u8; 11] = b"\x0Aic-request";
+const IC_STATE_ROOT_DOMAIN_SEPARATOR: &[u8; 14] = b"\x0Dic-state-root";
+const DER_PREFIX: &[u8; 37] = b"\x30\x81\x82\x30\x1d\x06\x0d\x2b\x06\x01\x04\x01\x82\xdc\x7c\x05\x03\x01\x02\x01\x06\x0c\x2b\x06\x01\x04\x01\x82\xdc\x7c\x05\x03\x02\x01\x03\x61\x00";
+const KEY_LENGTH: usize = 96;
+
+static INIT_BLS: Once = Once::new();
+
+// todo: do not merge until this is the actual Sodium key
```
p-shahi, line 45: ?

ic-agent/src/agent/mod.rs
```diff
@@ -428,6 +495,37 @@ impl Agent {
     }
 }
 
+fn initialize_bls() -> Result<(), AgentError> {
```
p-shahi, line 498: This file is getting big, does it make sense to move this, `extract_der`, and the lookup functions to a different file?

ic-agent/src/agent/agent_error.rs
```diff
@@ -78,6 +78,23 @@ pub enum AgentError {
...
+    #[error(
+        r#"BLS DER-encoded public key must be ${expected} bytes long, but is {actual} bytes long."#
+    )]
+    DerKeyLengthMismatch { expected: usize, actual: usize },
+
+    #[error("BLS DER-encoded public key is invalid. Expected the following prefix: ${expected:?}, but got ${actual:?}")]
+    DerPrefixMismatch { expected: Vec<u8>, actual: Vec<u8> },
+
+    #[error("The status response did not contain a root key")]
+    NoRootKeyInStatus(),
```
p-shahi, line 94: should we also print the `status` here (may be useful)

ic-agent/src/agent/mod.rs
```diff
@@ -136,9 +151,25 @@ impl Agent {
             ingress_expiry_duration: config
                 .ingress_expiry_duration
                 .unwrap_or_else(|| Duration::from_secs(300)),
+            root_key: RwLock::new(DEFAULT_ROOT_KEY.to_vec()),
         })
     }
 
+    /// Fetch the root key from the status endpoint.
+    /// It is not necessary to call this when communicating with "the" Internet Computer.
```
p-shahi, line 159: nit: maybe reword to affirm the positive "Only an agent communicating with a local/custom instance of the Internet Computer should call this method. It is not necessary to call this when communicating with the DFINITY foundation managed Internet Computer."

ic-agent/src/agent/mod.rs
```diff
@@ -136,9 +151,25 @@ impl Agent {
...
+    pub async fn fetch_root_key(&self) -> Result<(), AgentError> {
+        let status = self.status().await?;
+        let root_key = status.root_key.ok_or_else(AgentError::NoRootKeyInStatus)?;
+        let mut write_guard = self.root_key.write().unwrap();
+        *write_guard = root_key;
```
p-shahi, line 164:
```suggestion
        if let Ok(mut write_guard) = self.root_key.write() {
            *write_guard = root_key;
        }
```

ic-agent/src/agent/mod.rs
```diff
@@ -136,9 +151,25 @@ impl Agent {
...
+        Ok(())
+    }
+
+    fn read_root_key(&self) -> Result<Vec<u8>, AgentError> {
+        let root_key = self.root_key.read().unwrap().clone();
+        Ok(root_key)
```
p-shahi, line 170:
```suggestion
        if let Ok(read_lock) = self.root_key.read() {
            let root_key = *read_lock;
            Ok(root_key)
        } else {
            Err(AgentError::CouldntReadRootKey)
        }
```

ic-agent/src/agent/mod.rs
p-shahi: getting rid of the unwraps

ic-agent/src/agent/mod.rs
p-shahi: getting rid of unwraps and adding some new AgentError

ic-agent/src/agent/mod.rs
ericswanson-dfinity: Thanks!

ic-agent/src/agent/mod.rs
ericswanson-dfinity: Ok, I applied these changes.  The reason I had them as `unwrap()` is because the RwLock poisoning condition that would cause them to fail can't happen in the code as is, because the writer can't panic.

ic-agent/src/agent/mod.rs
ericswanson-dfinity: Good idea!

p-shahi (approved):

PR #87, opened 2020-11-24
fix: when content_type is missing, do not panic

hansl:


ic-agent/src/agent/agent_error.rs
```diff
@@ -104,7 +105,8 @@ impl HttpErrorPayload {
             } if is_plain_text_utf8(content_type) => {
                 f.write_fmt(format_args!(
                     "Http Error: status {}, content type {:?}, content: {}",
-                    status,
+                    StatusCode::from_u16(*status)
+                        .map_or_else(|_| format!("{}", status), |code| format!("{}", code)),
                     content_type.as_ref().unwrap(),
```
ericswanson-dfinity, line 110:
```suggestion
                    content_type.clone().unwrap_or("".to_string()),
```

ericswanson-dfinity (approved):
Thanks!

ic-agent/src/agent/agent_error.rs
hansl: First thing I tried. `&String` isn't `String`, unfortunately.
```
error[E0308]: mismatched types
   --> ic-agent/src/agent/agent_error.rs:126:53
    |
126 |                     content_type.as_ref().unwrap_or("".to_string()),
    |                                                     ^^^^^^^^^^^^^^
    |                                                     |
    |                                                     expected `&std::string::String`, found struct `std::string::String`
    |                                                     help: consider borrowing here: `&"".to_string()`
```

PR #90, opened 2020-12-05
feat: add support for identity based on an ECDSA key in an HSM

ericswanson-dfinity:
Adds `HardwareIdentity` in the `ic-identity-hsm` crate, which provides an Identity backed by an ECDSA key stored on a Hardware Security Module.


ic-identity-hsm/src/hsm.rs
```diff
@@ -0,0 +1,358 @@
+use ic_agent::{Identity, Signature};
+use ic_types::Principal;
+
+use num_bigint::BigUint;
```
hansl, line 4: Still needed?

ic-agent/Cargo.toml
```diff
@@ -22,16 +22,16 @@ hex = "0.4.0"
 ic-types = { path = "../ic-types", version = "0.1", features = [ "serde" ] }
 leb128 = "0.2.4"
 mime = "0.3.16"
-num-bigint = "0.2.6"
-openssl = "0.10.24"
+num-bigint = "0.3.1"
```
hansl, line 25: Still needed?

ic-agent/src/identity/basic.rs
```diff
@@ -3,7 +3,7 @@ use crate::{Identity, Signature};
 use num_bigint::BigUint;
```
hansl, line 3: Still needed?

ic-identity-hsm/src/hsm.rs
```diff
@@ -0,0 +1,358 @@
...
+const EXPECTED_EC_PARAMS: &[u8; 10] = b"\x06\x08\x2a\x86\x48\xce\x3d\x03\x01\x07";
+
+/// An error happened related to a HardwareIdentity.
+#[derive(Error, Debug)]
+pub enum HardwareIdentityError {
+    #[error(transparent)]
+    PKCS11(#[from] pkcs11::errors::Error),
+
+    #[error("ASN decode error {0}")]
+    ASN1Decode(ASN1DecodeErr),
```
hansl, line 36: Could you add a comment (not a documentation comment) that `ASN1DecodeError` does not implement the `Error` trait and so we cannot use `#[from]`? I'm sure someone will scratch their head about this in the future.

ic-identity-hsm/src/hsm.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+
+impl HardwareIdentity {
+    /// Create an identity using a specific key on an HSM.
+    /// The filename will be something like /usr/local/lib/opensc-pkcs11.s
+    /// The key_id must refer to a ECDSA key with parameters prime256v1 (secp256r1)
+    /// The key must already have been created.  You can create one with pkcs11-tool:
+    /// $ pkcs11-tool -k --slot $SLOT -d $KEY_ID --key-type EC:prime256v1 --pin $PIN
+    pub fn new<P>(
+        filename: P,
```
hansl, line 79: `filename` probably gives the wrong expectation here. What about something like `pkcs11_lib_path`?

ic-identity-hsm/src/hsm.rs
```diff
@@ -0,0 +1,358 @@
...
+    ctx: &Ctx,
+    session_handle: CK_SESSION_HANDLE,
+    pin: &str,
+    slot_id: CK_SLOT_ID,
+) -> Result<bool, HardwareIdentityError> {
+    let token_info = ctx.get_token_info(slot_id)?;
+    let login_required = token_info.flags & CKF_LOGIN_REQUIRED != 0;
+
+    if login_required {
+        ctx.login(session_handle, CKU_USER, Some(&pin))?;
```
hansl, line 141: Does this only work on command line prompts / terminals? What if I do `cat | dfx --identity hsm call ...`, will this fail or hang forever?

ic-identity-hsm/src/hsm.rs
```diff
@@ -0,0 +1,358 @@
...
+    Ok(der)
+}
+
+// Ensure that the key type is ECDSA.
+fn validate_key_type(
+    ctx: &Ctx,
+    session_handle: CK_SESSION_HANDLE,
+    object_handle: CK_OBJECT_HANDLE,
+) -> Result<(), HardwareIdentityError> {
+    // This value will be mutated.  `let mut` emits a warning, though.
```
hansl, line 181: is it mutated because of ffi? Could you add a note for "why" it's mutated and maybe wrap this in unsafe to make it double clear (this is super surprising to the reader and may lead to bugs).

hansl (changes requested):
Nothing much as a first pass. Will look more later.

ic-agent/Cargo.toml
ericswanson-dfinity: Yes, it's still needed, because the `oid!` macro uses BigUInt. (I tried to remove it)

ic-identity-hsm/src/hsm.rs
ericswanson-dfinity: Sounds good

ic-identity-hsm/src/hsm.rs
ericswanson-dfinity: There is no prompt / hang, because the PIN is passed as a parameter.  `dfx` will get it from environment variable `DFX_HSM_PIN`.

ic-identity-hsm/src/hsm.rs
ericswanson-dfinity: I don't know what `ffi` is.

ic-identity-hsm/src/hsm.rs
hansl: Ah I see. It sets the pin to the `&str` you gave. Okay. Not blocking this, but I think the flow is broken; from the PoV of this library we should ask the callee for the pin only if it's needed, but this flow requires it all the time. For dfx that makes sense, but another library could do a prompt for the PIN. So instead of always take a &str, we should take a `Fn() -> &str` and call it only when the pin is needed.

WDYT? Like I said, it's non blocking but if you want to improve.

ic-identity-hsm/src/hsm.rs
hansl: https://en.wikipedia.org/wiki/Foreign_function_interface

Basically bindings between libraries built in different languages. I think the whole pkcs11 library uses FFI to link against a library that's really in C, right?

ericswanson-dfinity: Maybe that is how it gets around it!

ic-identity-hsm/src/hsm.rs
ericswanson-dfinity: Sounds great, I will make this change

ic-identity-hsm/src/hsm.rs
```diff
@@ -129,16 +133,20 @@ fn open_session(
...
+fn login_if_required<PinFn>(
     ctx: &Ctx,
     session_handle: CK_SESSION_HANDLE,
-    pin: &str,
+    pin_fn: PinFn,
     slot_id: CK_SLOT_ID,
-) -> Result<bool, HardwareIdentityError> {
+) -> Result<bool, HardwareIdentityError>
+where
+    PinFn: FnOnce() -> Result<String, String>,
```
hansl, line 143: Just to make sure, this is only going to be called once, right?

hansl (approved):
LGTM! Thanks for the good work. The last 2 points are non blocking.

ic-identity-hsm/src/hsm.rs
ericswanson-dfinity: That's right.  We don't even hold onto a reference to the function.

PR #93, opened 2020-12-10
feat: Support AnonymousIdentity

### LINKED ISSUE #58: Make the default identity the anonymous user
Function `AgentBuilder::with_identity` has the following comment:

> This is required.

https://github.com/dfinity/agent-rs/blob/a921a935a4579f6e2c13a9a360795479be7f6c02/ic-agent/src/agent/builder.rs#L42

It should not be required. By default, it should use the anonymous user. The spec reads:
https://docs.dfinity.systems/public/#authentication

> The fields sender_pubkey and sender_sig must be omitted if and only if the sender field is absent or the anonymous principal.

nomeata:
and make it work (by not creating `sender_sig` and `sender_pubkey` fields)

This replaces the bogus `DummyIdentity`.

Also, remove the unused `principal` argument to `sign` (not sure what
the point was – doesn't the `Identity` _know_ who it is?)

This fixes #58


hansl:
> Also, remove the unused principal argument to sign (not sure what the point was – doesn't the Identity know who it is?)

Not necessarily. Basically, the API needs to stay stateless, but I wanted to allow switching identity per request. Should you rebuild the agent every time for e.g. a round robin use case? Building an agent can be expensive (e.g. querying the replica for a root key), so I decided it was simpler to pass the principal returned by the identity. This is just a pattern where the identity object is in control, but the API is stateless (because an agent might be doing multiple requests in parallel and the order of calls isn't guaranteed).

If you have an idea on how to tackle this, let me know.

ic-agent/src/identity/mod.rs
```diff
@@ -23,8 +23,6 @@ pub trait Identity: Send + Sync {
     fn sender(&self) -> Result<Principal, String>;
 
     /// Sign a blob, the concatenation of the domain separator & request ID,
-    /// creating the sender signature, with the principal passed in.
-    /// The principal should be
-    /// the same returned by the call to `sender()`.
-    fn sign(&self, blob: &[u8], principal: &Principal) -> Result<Signature, String>;
+    /// creating the sender signature.
+    fn sign(&self, blob: &[u8]) -> Result<Signature, String>;
```
hansl, line 27: This is a `BREAKING CHANGE` and should have a `# BREAKING CHANGE` section in the description.

hansl: Also see my comment above. I do have a use case :)

hansl:
Approving this. I just added some formatting.

hansl (approved):

luc-mercier:
Thanks!

Issue #91, opened 2020-12-09
Format HTTP errors with "text/html" content type as text


ericswanson-dfinity:
Or possibly text/* 

Example error that this would affect:
```
The replica returned an HTTP Error: Http Error: status 502 Bad Gateway,
content type "text/html", 
content: [60, 104, 116, 109, 108, 62, 13, 10, 60, 104, 101, 97, 100,
 62, 60, 116, 105, 116, 108, 101, 62, 53, 48, 50, 32, 66, 97, 100, 32, 
71, 97, 116, 101, 119, 97, 121, 60, 47, 116, 105, 
```

PR #97, opened 2020-12-18
feat: format any content-type: text as utf-8 text

ericswanson-dfinity:
Fixes https://github.com/dfinity/agent-rs/issues/91

For example, seen during a deploy:
```
± dfx ping ic
The replica returned an HTTP Error: Http Error: status 502 Bad Gateway,
content type "text/html",
content: [60, 104, 116, 109, 108, 62...]
```

### DIFF
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -179,7 +179,7 @@ fn format_as_text(content_type: &Option<String>) -> bool {
 
 #[cfg(test)]
 mod tests {
-    use crate::{HttpErrorPayload, AgentError};
+    use crate::{AgentError, HttpErrorPayload};
 
     #[test]
     fn http_payload_works_with_content_type_none() {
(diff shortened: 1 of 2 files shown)


ic-agent/src/agent/agent_error.rs
```diff
@@ -165,25 +165,75 @@ impl Display for HttpErrorPayload {
     }
 }
 
-fn is_plain_text_utf8(content_type: &Option<String>) -> bool {
-    // text/plain is also sometimes returned by the replica (or ic-ref),
-    // depending on where in the stack the error happens.
+fn format_as_text(content_type: &Option<String>) -> bool {
```
hansl, line 168: I would prefer to name this predicate as a true/false question; `is_text` or `can_format_as_text`.

This new name implies that it formats it.

hansl (approved):
Preemptively approving this with the note above. Love the new tests! Thanks

ic-agent/src/agent/agent_error.rs
ericswanson-dfinity: Good idea.  Thanks!

Issue #95, opened 2020-12-11
Add CI tests for HSM support


ericswanson-dfinity:
Use [SoftHSM](https://www.opendnssec.org/softhsm/) to test HardwareIdentity against ic-ref.

PR #104, opened 2021-01-14
test: Run tests using a software HSM

ericswanson-dfinity:
Adds a test phase that runs the ic-ref tests, but using the SoftHSM library to create identities rather than using basic identities.

This suite runs serially, because the SoftHSM library was panicking.

I also ran into trouble with two tests that create two identities.  The second call to `Ctx::new_and_initialize` panicked, so I made these two tests use a basic identity for the second identity.

This change surfaced an error related to slot ids.  The HSMs that I have used to date had slot index = slot id, but the SoftHSM library assigns different slot IDs.  So now we look up the slot id for the given slot index.

Updated to ic-ref 0.15.0, which added support for ECDSA keys.

Companion sdk PR: [PR-1316](https://github.com/dfinity/sdk/pull/1316)
Fixes https://github.com/dfinity/agent-rs/issues/95


p-shahi (approved):
lgtm!

PR #109, opened 2021-02-05
fix: use fresh expiry on request_status and read_state

p-shahi:


ic-agent/src/agent/mod.rs
```diff
@@ -634,11 +629,7 @@ impl<'agent> UpdateBuilder<'agent> {
         waiter.start();
 
         loop {
-            match self
-                .agent
-                .request_status_raw(&request_id, self.ingress_expiry_datetime)
```
ericswanson-dfinity, line 639: Is `self.ingress_expiry_datetime` something that we should remove altogether?

ericswanson-dfinity (approved):

ic-agent/Cargo.toml
```diff
@@ -36,6 +35,12 @@ thiserror = "1.0.20"
 url = "2.1.0"
 webpki-roots = "0.20.0"
 
+[dependencies.delay]
```
hansl, line 38: Woah let's not get this in master.

hansl (changes requested):

hansl (approved):

PR #120, opened 2021-03-03
feat: add an http-request interface and icx-http-server bin

hansl:
The icx-http-server is used to create a HTTP server which redirects every
requests to /api to the replica directly, and every other requests to an
http_request query call on the canister itself.

For the moment the design doc is private but will be added in this repo soon.

### DIFF
--- a/icx-http-server/Cargo.toml
+++ b/icx-http-server/Cargo.toml
@@ -14,7 +14,7 @@ hyper = { version = "0.14.4", features = ["full"] }
 ic-agent = { path = "../ic-agent" }
 ic-types = { path = "../ic-types" }
 ic-utils = { path = "../ic-utils" }
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1.2.0", features = ["full"] }
 serde = "1.0.115"
 serde_json = "1.0.57"
 url = "2.2.1"
--- a/icx-http-server/Cargo.toml
+++ b/icx-http-server/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "icx-http-server"
+version = "0.1.0"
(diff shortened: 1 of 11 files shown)


icx-http-server/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+candid = "0.6.15"
+clap = "3.0.0-beta.2"
+delay = "0.3.1"
+hyper = { version = "0.14.4", features = ["full"] }
+ic-agent = { path = "../ic-agent" }
+ic-types = { path = "../ic-types" }
+ic-utils = { path = "../ic-utils" }
+tokio = { version = "1", features = ["full"] }
```
ericswanson-dfinity, line 17: I wonder if we ought to specify the full version, so that we have to be explicit about getting new tokio versions:
```suggestion
tokio = { version = "1.2.0", features = ["full"] }
```

icx-http-server/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+) -> hyper::header::HeaderMap<hyper::header::HeaderValue> {
+    let mut result = hyper::HeaderMap::new();
+    for (k, v) in headers.iter() {
+        if !is_hop_header(k.as_str()) {
+            result.insert(k.clone(), v.clone());
+        }
+    }
+    result
+}
+use hyper::http::uri::Parts;
```
ericswanson-dfinity, line 159: Should this be at the top of the file with the other `use` statements?

ericswanson-dfinity (approved):
Looks good, just a couple small things

icx-http-server/src/main.rs
hansl: Yes, I was experimenting for a while.

PR #122, opened 2021-03-05
feat: add a ReplicaV1Facade trait to communicate with the replica

hansl:
# BREAKING CHANGE

This whole PR is a breaking change if you're using the previous API. The best you can do is use the `reqwest` feature and ignore the deprecation warning for `with_url`.

### DIFF
--- a/icx/src/main.rs
+++ b/icx/src/main.rs
@@ -3,8 +3,8 @@ use candid::types::{Function, Type};
 use candid::{check_prog, IDLArgs, IDLProg, TypeEnv};
 use clap::{crate_authors, crate_version, AppSettings, Clap};
 use ic_agent::agent::agent_error::HttpErrorPayload;
-use ic_agent::agent::http_facade::ReqwestHttpReplicaV1Facade;
-use ic_agent::agent::ReplicaV1Facade;
+use ic_agent::agent::http_transport::ReqwestHttpReplicaV1Transport;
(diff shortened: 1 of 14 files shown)


icx/src/main.rs
```diff
@@ -219,14 +233,80 @@ fn create_identity(maybe_pem: Option<PathBuf>) -> impl Identity {
...
+
+struct SerializingFacade;
+
+impl agent::ReplicaV1Facade for SerializingFacade {
+    fn read<'a>(
+        &'a self,
+        envelope: Vec<u8>,
+    ) -> Pin<Box<dyn Future<Output = Result<Vec<u8>, AgentError>> + Send + 'a>> {
+        async fn run(_: &SerializingFacade, envelope: Vec<u8>) -> Result<Vec<u8>, AgentError> {
+            print!("read\n\n{}", hex::encode(envelope));
```
ericswanson-dfinity, line 250: Are these `print` statements intentional and meant to not be `eprint`?

ericswanson-dfinity: (I see the separated println("result") / eprintln(result) below)

hansl: Yes, tthis facade is meant to serialize data to STDOUT, so it could be piped to a file or another process.

ericswanson-dfinity (approved):

PR #129, opened 2021-03-17
feat: add http_request_next method (large asset support)

ericswanson-dfinity:
Part of https://github.com/dfinity/dx-triage/issues/126

See also https://github.com/dfinity/sdk/pull/1512

### DIFF
--- a/ic-utils/src/canister.rs
+++ b/ic-utils/src/canister.rs
@@ -1,5 +1,6 @@
 use crate::call::{AsyncCaller, SyncCaller};
 use candid::de::ArgumentDecoder;
+use candid::parser::value::IDLValue;
 use candid::ser::IDLBuilder;
 use candid::CandidType;
 use ic_agent::{Agent, AgentError};
@@ -192,6 +193,25 @@ impl Argument {
         }
     }
 
+    /// Add an IDL Argument. If the current value of Argument is Raw, will set the
+    /// result to an error. If the current value is an error, will do nothing.
+    pub fn push_value_arg(&mut self, arg: IDLValue) {
+        match self.0 {
+            Ok(ArgumentType::Idl(ref mut idl_builder)) => {
+                let result = idl_builder.value_arg(&arg);
+                if let Err(e) = result {
+                    self.0 = Err(AgentError::CandidError(Box::new(e)))
+                }
+            }
+            Ok(ArgumentType::Raw(_)) => {
+                self.0 = Err(AgentError::MessageError(
+                    "Cannot overwrite a Raw Argument with a non-raw argument.".to_owned(),
+                ))
+            }
+            _ => {}
+        }
+    }
+
     /// Set the argument as raw, replacing any value that was there before. If the
     /// current argument was an error, does nothing.
     pub fn set_raw_arg(&mut self, arg: Vec<u8>) {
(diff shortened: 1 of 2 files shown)


ic-utils/src/interfaces/http_request.rs
```diff
@@ -29,6 +29,18 @@ pub struct HttpResponse {
...
+#[derive(CandidType)]
+pub struct HttpGetChunkRequest {
+    pub url: String,
+    pub content_encoding: String,
+    pub index: Nat,
+}
+
+#[derive(CandidType, Deserialize)]
+pub struct HttpGetChunkResponse {
+    pub chunk: Vec<u8>,
```
p-shahi, line 41:
```suggestion
    #[serde(with = "serde_bytes")]
    pub chunk: Vec<u8>,
```

p-shahi (approved):

ic-utils/src/interfaces/http_request.rs
```diff
@@ -29,6 +29,18 @@ pub struct HttpResponse {
     pub body: Vec<u8>,
 }
 
+#[derive(CandidType)]
+pub struct HttpGetChunkRequest {
```
p-shahi, line 33: can this be moved inside the body of `http_get_chunk` similar to https://github.com/dfinity/agent-rs/blob/next/ic-utils/src/interfaces/management_canister.rs#L453

hansl (changes requested):
Need design. See dfinity/dx-triage#126 where I'll start the discussion.

ic-utils/src/canister.rs
```diff
@@ -192,6 +193,25 @@ impl Argument {
         }
     }
 
+    /// Add an IDLValue Argument. If the current value of Argument is Raw, will set the
+    /// result to an error. If the current value is an error, will do nothing.
+    pub fn push_value_arg(&mut self, arg: IDLValue) {
```
hansl, line 198: `IDLValue` implements CandidType, so you should be able to use `push_idl_arg` with an `IDLValue`. Was there any issue with using that other function?

ic-utils/src/canister.rs
```diff
@@ -258,6 +278,16 @@ impl<'agent, 'canister: 'agent, Interface> SyncCallBuilder<'agent, 'canister, In
         self
     }
 
+    /// Add an argument to the candid argument list. This requires Candid arguments, if
+    /// there is a raw argument set (using [with_arg_raw]), this will fail.
+    pub fn with_value_arg(
```
hansl, line 283: Same question as above.

hansl (changes requested):

ic-utils/src/canister.rs
ericswanson-dfinity: Yes, `IDLValue::ty()` calls unreachable() : https://github.com/dfinity/candid/blob/master/rust/candid/src/parser/value.rs#L326
called from here: https://github.com/dfinity/candid/blob/master/rust/candid/src/ser.rs#L28

hansl (changes requested):
.

hansl (approved):
Okay LGTM, but this is an anti-pattern. We might have to deprecate these methods if we make IDLValue work.

ericswanson-dfinity:
Entered https://github.com/dfinity/agent-rs/issues/132 to make `with_arg` work with `IDLValue`.

hansl (approved):

PR #133, opened 2021-03-19
feat: support HTTP Request streaming in icx-http-request

hansl:


hansl:
This is a draft until I can get a ic-ref test.

icx-http-server/src/main.rs
```diff
@@ -116,19 +120,58 @@ async fn forward_request(
...
+        let agent = agent.as_ref().clone();
+        sender.send_data(Bytes::from(http_response.body)).await?;
+
+        tokio::spawn(async move {
+            let canister = HttpRequestCanister::create(&agent, canister_id);
+            // We already call it once.
+            let mut count = 1;
+            loop {
+                count += 1;
+                if count >= MAX_HTTP_REQUEST_NEXT_CALL_COUNT {
```
ericswanson-dfinity, line 146: If `MAX_HTTP_REQUEST_NEXT_CALL_COUNT` means "max number of calls to `http_request_next`, then:
```suggestion
            // We have not yet called http_request_next.
            let mut count = 0;
            loop {
                count += 1;
                if count > MAX_HTTP_REQUEST_NEXT_CALL_COUNT {
```

Example: if `MAX_HTTP_REQUEST_NEXT_CALL_COUNT` were 0, this would allow 0 calls to `http_request_next`, if 1, then 1, if 2, then 2.

As written, if `MAX_HTTP_REQUEST_NEXT_CALL_COUNT` were 2, this would allow 0 calls to `http_request_next`, and 1 chunk of data total, because `count` would be 2 on the first `>=` check.

hansl: Good call.. That will show me to code late...

PR #135, opened 2021-03-26
fix: always parse HttpError content as UTF-8 whatever the content type

lwshang:
The replica often returns HttpError in which the content type is empty but the content is actually UTF-8 String. Before, people had to manually convert the byte array into String to see the meaningful content. With this fix, the content will always be parsed whatever the content type is. 

We should also check the replica implementation why the content type was not set correctly.


ericswanson-dfinity (approved):

lwshang:
@hansl Could you merge this PR?

hansl (approved):
Approving this, but there's too much cloning happening. This can be done without any new memory copy.

PR #143, opened 2021-04-13
chore: fix cabal cache key in CI by including project hash

nomeata:
the Github Actions configuration for the Cabal cache was not good
(despite it being the recommended configuration on the cache action
documentation; it’s just broken). The cache key is too fixed, so if
`ic-ref` changes, you still get a cache hit with the old store, and
still rebuild lots of dependendencies, but don’t push a new cache.

This tries to improve this, by including some of cabals configuration
files in the cache key.

(This is the same as https://github.com/dfinity/agent-js/pull/326)


.github/workflows/ic-ref.yml
```diff
@@ -39,7 +39,9 @@ jobs:
         with:
           path: |
             ~/.cabal/store
-          key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-store
+          key: cabal-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('ic-ref/impl/cabal.project', 'ic-ref/impl/cabal.project.freeze') }}
```
ericswanson-dfinity, line 42:
```suggestion
          key: cabal-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('ic-hs/cabal.project', 'ic-hs/cabal.project.freeze') }}
```

ericswanson-dfinity (approved):

PR #145, opened 2021-04-14
feat: add verbose logging to icx-proxy

hansl:


icx-proxy/src/logging.rs
```diff
@@ -0,0 +1,79 @@
...
+        // A Tee mode is basically 2 drains duplicated.
+        LoggingMode::Tee(out) => Logger::root(
+            slog::Duplicate::new(
+                create_drain(LoggingMode::Stderr),
+                create_drain(LoggingMode::File(out)),
+            )
+            .fuse(),
+            slog::o!(),
+        ),
+        _ => {
```
ericswanson-dfinity, line 37: Suggestion: be explicit about the other case so that a compiler error will result if an entry is added to the enum later, without also adding a handler case here.
```suggestion
        LoggingMode::Stderr => {
```

icx-proxy/src/logging.rs
```diff
@@ -0,0 +1,79 @@
...
+
+pub(crate) fn setup_logging(opts: &Opts) -> Logger {
+    // Create a logger with our argument matches.
+    let verbose_level = opts.verbose as i64 - opts.quiet as i64;
+    let logfile = opts.logfile.clone().unwrap_or("log.txt".into());
+
+    let mode = match opts.logmode.as_str() {
+        "tee" => LoggingMode::Tee(logfile),
+        "file" => LoggingMode::File(logfile),
+        _ => LoggingMode::Stderr,
```
ericswanson-dfinity, line 53: `clap` will keep this from happening unless a new logmode is added, but better to get a failure message than to silently keep logging to stderr.
```suggestion
        "stderr" => LoggingMode::Stderr,
        _ => unreachable!("unhandled logmode"),
```

icx-proxy/src/main.rs
```diff
@@ -131,7 +183,7 @@ async fn forward_request(
         builder = builder.header(&name, value);
     }
 
-    if let Some(streaming_strategy) = http_response.streaming_strategy {
+    let (body, response) = if let Some(streaming_strategy) = http_response.streaming_strategy {
```
ericswanson-dfinity, line 186: I think we can use the same logic for the `body` in the streaming and non-streaming case, because the logging code is only going to display the first few bytes anyway:

```suggestion
    let body = if logger.is_trace_enabled() {
        Some(http_response.body.clone())
    } else {
        None
    };
    let response = if let Some(streaming_strategy) = http_response.streaming_strategy {
```

hansl: Right, but I'll change the code a bit to be more explicit here.

icx-proxy/src/main.rs
```diff
@@ -187,10 +239,51 @@ async fn forward_request(
...
+
+        slog::trace!(logger, ">>");
+        slog::trace!(
+            logger,
+            ">> {}{}",
+            match std::str::from_utf8(&body) {
+                Ok(s) => format!(r#""{}""#, s[..usize::min(100, s.len())].escape_default()),
+                Err(_) => hex::encode(&body[..usize::min(100, body.len())]),
+            },
+            if body.len() > 500 {
```
ericswanson-dfinity, line 278: Could we use the same cutoff here as we are using for the slice size above?
```suggestion
            if body.len() > 100 {
```

hansl: Yeah and I should make it a constant.

icx-proxy/src/main.rs
ericswanson-dfinity: or something like
```
let (first_part_of_body, total_body_length) = if logger.is_trace_enabled() {
    (Some(body[..usize::min(100, body.len()), Some(body.len()))
} else {
    (None, None)
}
```

ericswanson-dfinity (approved):

icx-proxy/src/main.rs
```diff
@@ -187,10 +248,49 @@ async fn forward_request(
...
+
+        slog::trace!(logger, ">>");
+        slog::trace!(
+            logger,
+            ">> {}{}",
+            match std::str::from_utf8(&body) {
+                Ok(s) => format!(
+                    r#""{}""#,
+                    s[..usize::min(MAX_LOG_BODY_SIZE, s.len())].escape_default()
+                ),
```
ericswanson-dfinity, line 280: Is `format` necessary here?
```suggestion
                Ok(s) => s[..usize::min(MAX_LOG_BODY_SIZE, s.len())].escape_default(),
```

Issue #149, opened 2021-04-18
[icx-proxy] Add a list of canister ID aliases


hansl:
The `identity` service team is asking that they use `identity.ic0.app` as their DNS. This resolves to a specific canister ID. This means that the `icx-proxy` needs to be able to resolve this DNS to their canister.

# Proposed Design

Add a `--dns-alias identity.ic0.app:aaaaa-aa` command line argument that maps domain names to canister IDs. This would be checked first before using any other resolution logic. Any additional subdomains (in this example, `abc.identity.ic0.app`) would also match.

This command line argument can be repeated to add more aliases.

hansl:
Additional: I'm okay with either `:` or `=` as separators between domain names and canisterId in the argument value.

PR #150, opened 2021-04-19
feat: [icx-proxy] Add a list of canister ID aliases

ericswanson-dfinity:
Adds a `--dns-alias identity.ic0.app:aaaaa-aa` command line argument that maps domain names to canister IDs.

Implements https://github.com/dfinity/agent-rs/issues/149


icx-proxy/src/main.rs
```diff
@@ -69,13 +71,27 @@ pub(crate) struct Opts {
...
+    dns_aliases: &DnsAliases,
+) -> Option<Principal> {
     let url = Uri::from_str(hostname).ok()?;
 
     // Check if it's localhost or ic0.
-    match url.host()?.split('.').collect::<Vec<&str>>().as_slice() {
+    let host_parts = url.host()?.split('.').collect::<Vec<&str>>();
+    let host_parts = host_parts.as_slice();
+
+    if let Some(principal) = dns_aliases.resolve_canister_id_from_host_parts(host_parts) {
```
hansl, line 91:
```suggestion
    if let Some(principal) = dns_aliases.resolve_canister_id_from_host_parts(&host_parts) {
```

icx-proxy/src/dns_aliases.rs
```diff
@@ -0,0 +1,67 @@
...
+    // dns_suffix_slice2: &[&str],
+    principal: Principal,
+}
+
+#[derive(Clone, Debug)]
+pub struct DnsAliases {
+    dns_aliases: Vec<DnsAlias>,
+}
+
+impl DnsAliases {
```
hansl, line 27: Add some comments here and for the pub methods of this struct.

icx-proxy/src/config/canister_dns_config.rs
```diff
@@ -0,0 +1,150 @@
...
+
+impl CanisterDnsConfig {
+    /// Create a CanisterDnsConfig instance from command-line configuration.
+    /// dns_aliases: 0 or more entries of the form of dns.alias:canister-id
+    pub fn new(dns_aliases: &[String]) -> anyhow::Result<CanisterDnsConfig> {
+        let dns_aliases = dns_aliases
+            .iter()
+            .map(|alias| DnsAlias::new(alias))
+            .collect::<anyhow::Result<Vec<DnsAlias>>>()?;
+        Ok(CanisterDnsConfig { dns_aliases })
```
hansl, line 18: I filed https://github.com/dfinity/agent-rs/issues/153 which is non blocking for this but would solve a gotcha when passing subdomains as aliases.

hansl (approved):
One non blocking comment, if you fix it here that's great please mark the bug as fixed too.

Thanks for the good work!

nomeata:
BTW, did you spot that we were using this feature (off this branch) in the Demo today, when Christoph browsed to `https://identity.localhost`?

paulyoung:
@kritzcreek, could you help me understand how you were able to get `name.localhost` to work?

I tried `icx-proxy --dns-alias name.localhost:rrkah-fqaaa-aaaaa-aaaaq-cai` but requests made to it via `curl` result in `Could not resolve host: name.localhost`

Did you just add an entry to `/etc/hosts`, or something else?

kritzcreek:
@paulyoung IIRC I just followed these steps at the time: https://github.com/dfinity/internet-identity/blob/cb75f38160ee60cfb0ab986b5cbabfff00d21e5e/Demo.md#setup

I definitely didn't mess with the hosts file.

paulyoung:
Can anyone else help me understand how this is intended to work?

I think I followed the same steps as above but I can’t see how anything is supposed to be servicing requests at `x.localhost` domains.

hansl:
`curl` could work if you use the `Host:` header and use `localhost` instead of trying to resolve a subdomain (which doesn't work in standard tooling).

paulyoung:
Thanks @hansl

PR #152, opened 2021-04-20
chore: checkout the right SHA on CI on pull_request_target

hansl:


.github/workflows/ic-ref.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           path: main
+          ref: ${{ github.event.pull_request.head.sha }}
```
ericswanson-dfinity, line 30: I don't think this will be available on `push` events

ericswanson-dfinity:
`pull_request_target` removed from github workflow

Issue #153, opened 2021-04-21
[icx-proxy] Sort aliases from longest to shortest


hansl:
DNS Aliases should be compared from longest to shortest, as `a.b.c` should resolve even if we also resolve `b.c` to something else, if both a passed as aliases.

PR #155, opened 2021-04-21
fix: [icx-proxy] Sort aliases from longest to shortest

ericswanson-dfinity:
Fixes https://github.com/dfinity/agent-rs/issues/153

### DIFF
--- a/icx-proxy/src/config/canister_dns_config.rs
+++ b/icx-proxy/src/config/canister_dns_config.rs
@@ -18,7 +18,6 @@ impl CanisterDnsConfig {
             .iter()
             .map(|alias| DnsAlias::new(&alias))
             .collect::<anyhow::Result<Vec<DnsAlias>>>()?;
-
         Ok(CanisterDnsConfig { dns_aliases })
     }
 
@@ -190,7 +189,7 @@ mod tests {
     #[test]
     fn searches_longest_to_shortest_not_alpha() {
         // Similar to searches_longest_to_shortest, but make sure we
-        // are sorting by length, not alphabetically
+        // don't happen to get there by sorting alphabetically
         let dns_aliases = parse_dns_aliases(vec![
             "a.x.c:r7inp-6aaaa-aaaaa-aaabq-cai",
             "x.c:rrkah-fqaaa-aaaaa-aaaaq-cai",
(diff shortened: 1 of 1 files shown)


icx-proxy/src/config/canister_dns_config.rs
```diff
@@ -11,9 +12,11 @@ impl CanisterDnsConfig {
     /// Create a CanisterDnsConfig instance from command-line configuration.
     /// dns_aliases: 0 or more entries of the form of dns.alias:canister-id
     pub fn new(dns_aliases: &[String]) -> anyhow::Result<CanisterDnsConfig> {
+        let mut dns_aliases: Vec<String> = dns_aliases.iter().map(String::clone).collect();
+        dns_aliases.sort_by_key(|s| Reverse(s.len()));
         let dns_aliases = dns_aliases
             .iter()
-            .map(|alias| DnsAlias::new(alias))
+            .map(|alias| DnsAlias::new(&alias))
             .collect::<anyhow::Result<Vec<DnsAlias>>>()?;
```
hansl, line 20: I with we wouldn't have to collect twice, but oh well... It's just at startup anyway.

icx-proxy/src/config/canister_dns_config.rs
```diff
@@ -143,6 +146,86 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn searches_longest_to_shortest_not_alpha_reversed() {
+        // Similar to searches_longest_to_shortest, but make sure we
+        // don't happen to get there by sorting alphabetically/reversed
+        let dns_aliases = parse_dns_aliases(vec![
+            "x.a.c:r7inp-6aaaa-aaaaa-aaabq-cai",
+            "a.c:rrkah-fqaaa-aaaaa-aaaaq-cai",
```
hansl, line 215: Can you reverse the order you pass them? If it's not sorted this test would still pass, right?

hansl (changes requested):
One non blocking, and another to make the test better. Otherwise LGTM

icx-proxy/src/config/canister_dns_config.rs
ericswanson-dfinity: The test above passes them in the reversed order from this one.  Does that cover the case you were wanting to?

icx-proxy/src/config/canister_dns_config.rs
ericswanson-dfinity: Me too, but I couldn't sort in between `.iter()` and `.collect()`..  I have an idea though

icx-proxy/src/config/canister_dns_config.rs
ericswanson-dfinity: Yeah, I missed something on this one.. will adjust

hansl (approved):

PR #161, opened 2021-04-30
Add Cargo.lock

ninegua:
`icx` turns out to be a useful tool that we can use for other projects (for testing purpose, e.g.).

Adding Cargo.lock makes it easier to build as a dependency.


chenyan-dfinity:
Have you tried `ic-repl`: https://github.com/chenyan2002/ic-repl.

We've been using it for tests in CanCan: https://github.com/dfinity/cancan/blob/main/service/tests/viral.test.sh

ninegua:
Turns out to be a bigger change, closing this.

PR #162, opened 2021-04-30
fix: Make agent-rs compatibile with dfinity repo

ninegua:
The goal is to use icx or icx-proxy with dfinity repo as a separately compiled Rust package.

Because we are still using `rustc` 1.45.2 in the dfinity repo, which is incompatible with socket2 0.4.0. To work around this problem, we have to downgrade a bunch of dependencies, eventually resulted in moving from tokio 1.2.0 to 0.2.25.


chenyan-dfinity:
I had the same problem bumping Candid in dfinity repo: https://github.com/dfinity-lab/dfinity/pull/10300

I think the better solution is to bump rustc in dfinity?

ninegua:
> I think the better solution is to bump rustc in dfinity?

Yes, I agree.

ninegua:
All tests pass, and ic-types are not touched. I think this is ready to merge.

hansl (changes requested):
Can I get an explanation of why we would reduce our Tokio and Reqwest versions? Tokio is more notable here as we were on the stable post 1.0 release, and this is a major step back.

hansl:
I'm way more okay if you want to have a "monorepo compatible" branch where this is applied. But we publish this on crates.io and there will be more people than just the IC itself who uses it.

ninegua:
> Can I get an explanation of why we would reduce our Tokio and Reqwest versions?

The explanation was in the description. Essentially a cascading effect of having to downgrade socket2 from 0.4.0. 

> I'm way more okay if you want to have a "monorepo compatible" branch where this is applied

So @ericswanson-dfinity 's request is to use icx-proxy in sdk. Are you ok that sdk depends on this branch instead the "next" branch?

hansl:
> So @ericswanson-dfinity 's request is to use icx-proxy in sdk. Are you ok that sdk depends on this branch instead the "next" branch?

Yes.

PR #164, opened 2021-05-04
feat: use hardcoded IC public key by default

ericswanson-dfinity:
Also added `--fetch-root-key` option to `icx`

To get the value for the root key, I ran this through a base64 decoder
```
-----BEGIN PUBLIC KEY-----
MIGCMB0GDSsGAQQBgtx8BQMBAgEGDCsGAQQBgtx8BQMCAQNhAIFMDm7HH6tYOwi9
gTc8JVw8NxsuhIY8mKTx4It0I10U+12cDNVG2WhfkToMCyzFNBWDv0tDkuRn25bW
W5u0y3FxEvhHLg1aTRRQX/10hLASkQkcX4e5iINGP5gJGguqrg==
-----END PUBLIC KEY-----
```

See also https://github.com/dfinity/sdk/pull/1643

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -198,13 +198,14 @@ impl Agent {
         self.transport = Arc::new(transport);
     }
 
-    /// Fetch the root key of the replica using its status end point, and update the agent's
-    /// root key. This only uses the agent's specific upstream replica, and does not ensure
-    /// the root key validity. In order to prevent any MITM attack, developers should try
-    /// to contact multiple replicas.
+    /// By default, the agent is configured to talk to the main Internet Computer, and verifies
+    /// responses using a hard-coded public key.
     ///
-    /// The root key is necessary for validating state and certificates sent by the replica.
-    /// By default, it is set to [None] and validating methods will return an error.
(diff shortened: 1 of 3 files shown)


nomeata (commented):
The docs currently says
```
    /// Fetch the root key of the replica using its status end point, and update the agent's
    /// root key. This only uses the agent's specific upstream replica, and does not ensure
    /// the root key validity. In order to prevent any MITM attack, developers should try
    /// to contact multiple replicas.
    ///
    /// The root key is necessary for validating state and certificates sent by the replica.
    /// By default, it is set to [None] and validating methods will return an error.
```

I’d propose to change this to
```
    /// By default, the agent is configured to talk to the main Internet Computer, and verifies
    /// responses using a hard-coded public key.
    /// 
    /// This function will instruct the agent to ask the endpoint for its public key, and use
    /// that instead. This is required when talking to a local test instance, for example.
    ///
    /// *Only use this when you are  _not_ talking to the main Internet Computer, else
    /// you are prone to man-in-the-middle attacks! Do not call this function by default.*
```
or something like that.

p-shahi (approved):

PR #166, opened 2021-05-06
fix: icx-proxy should only consult query string and referrer when run locally

ericswanson-dfinity:
Only check the URI (query string) and `referer` header when `icx-proxy` is run for local development, determined by a new `--local` command-line argument.

### DIFF
--- a/icx-proxy/src/main.rs
+++ b/icx-proxy/src/main.rs
@@ -81,6 +81,10 @@ pub(crate) struct Opts {
     /// is used as the Principal, if it parses as a Principal.
     #[clap(long, default_value = "localhost")]
     dns_suffix: Vec<String>,
+
+    /// If true, check the URI and the referer when determining canister id.
+    #[clap(long)]
+    check_uri_and_referer_for_canister_id: bool,
 }
 
 fn resolve_canister_id_from_hostname(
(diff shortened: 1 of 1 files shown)


nomeata:
Thanks!

icx-proxy/src/main.rs
```diff
@@ -465,7 +461,9 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let counter = AtomicUsize::new(0);
     let debug = opts.debug;
-    let check_uri_and_referer_for_canister_id = opts.check_uri_and_referer_for_canister_id;
+
+    /// For local/development, check the URI (query string) and the referer for a canister id.
+    let check_uri_and_referer_for_canister_id = opts.address.ip().is_loopback();
```
nomeata: Nice in principle. But does that work? Are we sure that the deployed icx-proxy doesn’t sit behind some form of proxy _on the same machine_?

ericswanson-dfinity: Yeah, it does not work, both because of that and because in production icx-proxy binds to localhost, behind nginx

nomeata: Yes, that’s what what I mean. Too bad.

ericswanson-dfinity: Hh yes that is exactly what you wrote

ericswanson-dfinity:
icx-proxy moved to https://github.com/dfinity/icx-proxy

PR #172, opened 2021-05-13
chore: upgrade to candid 0.6.21

ninegua:
This should be stacked on top of #162


Daniel-Bloom-dfinity:
Upgraded to `0.7` so dropping this.

PR #173, opened 2021-05-13
fix: Make agent-rs compatible with dfinity repo and sdk under nix

ericswanson-dfinity:
Makes it possible to build icx and icx-proxy with rust 1.45.2, which is what the sdk repo uses to build under nix.

Patched from https://github.com/ninegua/agent-rs/tree/paulliu/downgrade
See also https://github.com/dfinity/agent-rs/pull/162

This PR will be held open, not merged, and otherwise kept up-to-date with `next`.  Its purpose is to allow the sdk repo to build `icx` and `icx-proxy` under nix.

The intent is to close this PR without merging after the `dfinity` repo is updated to build with at least rust 1.47.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.15"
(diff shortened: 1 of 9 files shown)


ericswanson-dfinity:
renamed branch dfx-support/...

PR #176, opened 2021-05-17
refactor: UpdateCall, Agent.poll, Agent.wait

ericswanson-dfinity:
refactor Agent.call_and_wait so that `call` and `wait` functionality can be called separately, and added a `poll` method

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -39,6 +39,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
+use std::task::{Context, Poll};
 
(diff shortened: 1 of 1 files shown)


hansl:
@ericswanson-dfinity this _looks_ ready to review, but still marked as draft. Should we make it a proper PR and start the review?

ic-agent/src/agent/mod.rs
```diff
@@ -408,6 +422,75 @@ impl Agent {
...
+                            .restart()
+                            .map_err(|_| AgentError::WaiterRestartError())?;
+                        request_accepted = true;
+                    }
+                }
+                PollResult::Completed(result) => return Ok(result),
+            };
+
+            waiter
+                .wait()
```
hansl, line 489: `async_wait` with Garcon, now.

hansl (changes requested):
On change otherwise LGTM.

hansl (approved):

PR #180, opened 2021-05-18
refactor: move to using Garcon which supports async

hansl:


chenyan-dfinity (approved):

ericswanson-dfinity (approved):
I do think we should specify `major.minor.patch` in dependencies for reproducible builds, but I will defer to you on this.

hansl:
@ericswanson-dfinity good idea, I'll do that for the binaries, but not the libraries as they should be more relaxed in what they accept.

PR #182, opened 2021-05-22
chore: make function/structs public for sdk#1676

lwshang:
dfinity/sdk#1676

In order to handle response more user friendly in `sign/send`.

### DIFF
--- a/ic-agent/src/agent/replica_api.rs
+++ b/ic-agent/src/agent/replica_api.rs
@@ -104,7 +104,7 @@ pub struct ReadStateResponse {
 
 /// A `Certificate` as defined in https://docs.dfinity.systems/public/#_certificate
 #[derive(Deserialize)]
-pub(crate) struct Certificate {
+pub struct Certificate {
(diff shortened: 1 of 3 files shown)


p-shahi (approved):

ericswanson-dfinity (dismissed):

ic-agent/src/agent/replica_api.rs
```diff
@@ -104,7 +104,7 @@ pub struct ReadStateResponse {
 
 /// A `Certificate` as defined in https://docs.dfinity.systems/public/#_certificate
 #[derive(Deserialize)]
-pub(crate) struct Certificate {
+pub struct Certificate {
```
hansl, line 107: We don't want to expose anything from replica_api. Any structure in this file needs to stay private. That's because they are ADT, not proper user facing APIs. 

You should build an external facing Certificate which includes the methods you want.

hansl (changes requested):

PR #183, opened 2021-05-24
refactor: move HashTree to use Cows instead of owning the values

hansl:
This allows us to reduce memory and processing time. This is yak shaving for
publishing this data structure in ic_types and making it WASM (and canister)
friendly, with the hope we can move the certified asset canister to using this
(and hopefully more people).


hansl:
I've added Roman to the review as he's the one who implemented the asset canister with HashTree for certification.

My next steps are to make this public in `ic_types`, then implement a `icx-cert` that validates certifications from a query call.

ericswanson-dfinity (approved):

roman-kashitsyn (approved):

PR #191, opened 2021-05-26
feat: --proxy: forward /_/ to a specified url

ericswanson-dfinity:
Add command-line parameter `--proxy <url>` which, if present, forwards requests with path beginning with `/_/` to the specified endpoint.

Used by `dfx` to serve data from  `/_/candid`

### DIFF
--- a/icx-proxy/src/main.rs
+++ b/icx-proxy/src/main.rs
@@ -418,7 +418,7 @@ async fn forward_api(
     Ok(response)
 }
 
-async fn not_found() -> Result<Response<Body>, Box<dyn Error>> {
+fn not_found() -> Result<Response<Body>, Box<dyn Error>> {
     Ok(Response::builder()
         .status(StatusCode::NOT_FOUND)
         .body("Not found".into())?)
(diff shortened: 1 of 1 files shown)


icx-proxy/src/main.rs
```diff
@@ -418,17 +422,26 @@ async fn handle_request(
...
-    match if request.uri().path().starts_with("/api/") {
+    let request_uri_path = request.uri().path();
+    match if request_uri_path.starts_with("/api/") {
         slog::debug!(
             logger,
             "URI Request to path '{}' being forwarded to Replica",
             &request.uri().path()
         );
         forward_api(&ip_addr, request, &replica_url).await
+    } else if proxy_url.is_some() && request_uri_path.starts_with("/_/") {
```
hansl, line 438: Need to return 404 if there's no `proxy_url` configured.

hansl: Ah I see we didn't do that before. I won't block this PR on it then.

ericswanson-dfinity: I'm also not clear on how this is supposed to work outside of `dfx`

ericswanson-dfinity: Anyway, this is updated to return 404 now

hansl (approved):
LGTM. Loves adding 32 lines to make this happen! This is awesome

Issue #190, opened 2021-05-26
Make more `Principal` functions `const`


Daniel-Bloom-dfinity:
Now that `Principal` is no longer heap based, more of it can be made `const`.

PR #192, opened 2021-05-27
feat: make more `Principal` functions `const` (#190)

Daniel-Bloom-dfinity:
Also some minor comment cleanup

### DIFF
--- a/ic-types/src/principal.rs
+++ b/ic-types/src/principal.rs
@@ -3,6 +3,22 @@ use std::convert::TryFrom;
 use std::fmt::Write;
 use thiserror::Error;
 
+// TODO: remove (blocked by rust-lang/rust#85194)
+const ASSERT: [(); 1] = [()];
+macro_rules! const_panic {
+    ($($arg:tt)*) => {
+        #[allow(unconditional_panic)]
+        let _ = $crate::principal::ASSERT[1];
+    };
+}
+
+/// An error happened while decoding bytes to make a principal.
+#[derive(Error, Clone, Debug, Eq, PartialEq)]
+pub enum PrincipalBufferError {
+    #[error("Buffer is too long.")]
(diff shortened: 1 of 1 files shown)


ic-types/src/principal.rs
```diff
@@ -136,8 +161,35 @@ impl Principal {
...
+        ]))
+    }
+
+    /// Attempt to decode a slice into a Principal. `panic`s if the bytes can't be interpreted.
+    pub const fn from_slice(bytes: &[u8]) -> Self {
+        if let Ok(v) = Self::try_from_slice(bytes) {
+            v
+        } else {
+            // TODO: replace with panic (blocked by rust-lang/rust#85194)
+            const_panic!("slice length exceeded capacity");
```
hansl, line 176: Does this translate into a regular `panic!()` when the function is called as a non-const?

ic-types/src/principal.rs
```diff
@@ -351,6 +406,67 @@ mod inner {
...
+        pub const fn new() -> Self {
+            PrincipalInner {
+                len: 0,
+                bytes: [0; Self::MAX_LENGTH_IN_BYTES],
+            }
+        }
+
+        /// Panics if the length is over `HASH_LEN_IN_BYTES`
+        ///
+        /// ```compile_fail
```
hansl, line 418: I still would prefer to test those methods on `Principal` itself to avoid this hidden structure.

What I'm thinking of:
1. test compile-time behaviour on Principal,
2. test same runtime behaviour on PrincipalInner?

Since `Principal` is what's exported, it's okay to test `const` behaviour on it. WDYT?

Daniel-Bloom-dfinity: Sounds reasonable to me.

hansl (approved):
LGTM. Good work! This looks awesome.

PR #199, opened 2021-06-04
feat: add `Copy` to `Principal` and serde to `PrincipalError`

Daniel-Bloom-dfinity:


ic-types/src/principal.rs
```diff
@@ -187,10 +190,11 @@ impl Principal {
 
     /// Attempt to decode a slice into a Principal.
     pub const fn try_from_slice(bytes: &[u8]) -> Result<Self, PrincipalError> {
+        const ANONYMOUS: u8 = PrincipalClass::Anonymous as u8;
```
hansl, line 193: Nice :+1: thanks!

hansl (approved):
LGTM. That's awesome!

PR #200, opened 2021-06-04
feat: support sign/send in `Agent`

lwshang:
### New Public APIs
* `QueryBuilder` and `UpdateBuilder` now have `sign` method which generates the signed query/update in CBOR encoding.
* `UpdateBuilder` has `sign_request_status` method which generated the signed the request_status for the corresponding update.
* `Agent` has `query_signed`, `update_signed` and `request_status_signed`. They respectively send out the signed CBOR bytes and handle the response in the same way as their "raw" counterparts, i.e. `query_raw`, `update_raw`, `request_status_raw`.

### Discussion
My current design/implementation requires the user to provide all fields of the request together with the bytes. So that in these `*_signed` methods, the fields can be checked with the value in CBOR bytes and abort if mismatch.
Should we extract the check parts as separate methods? For example, we can have `check_signed_query` only do the check and `query_signed` send the bytes as long as it is a valid query (can be successfully deserialized). Therefore, users can opt in to inspect CBOR bytes depending on trust of source.


lwshang:
### Recent change
I add public structs `SignedQuery`, `SignedUpdate` and `SignedRequestStatus` for interchanging the signed CBOR bytes together with the necessary fields data to validate and send.

All these structs  `#[derive(Serialize, Deserialize)]`, so that they can be reliably sent between machines in json files or other formats. As a dependency, `RequestId` now has to be `#[derive(Serialize, Deserialize)]`.

Instead of return only the signed CBOR bytes, `QueryBuilder::sign()`, `UpdateBuilder::sign()` and `UpdateBuilder::sign_request_status` now returns the new structs above respectively.

As said in the "Discussion" section of the first PR comment, I extracted the code for validating the CBOR bytes into separate functions: `signed_query_inspect`, `signed_update_inspect` and `signed_request_status_inspect`. And `query_signed`, `update_signed` and `request_status_signed` takes less arguments since they no longer inspect the CBOR bytes.

lwshang:
### Summary of new APIs

1. `ic-agent/src/agent/signed.rs`
```rust
pub struct SignedQuery;
pub struct SignedUpdate;
pub struct SignedRequestStatus;
```
These three public structs are added for exchanging the signed CBOR bytes with necessary field data.

2. `ic-agent/src/agent/mod.rs`
```rust
impl UpdateBuilder {
    pub fn sign;
}
impl UpdateBuilder {
    pub fn sign;
}
impl Agent {
    pub fn sign_request_status;
}
```
These three methods generate the corresponding structs in point 1.

3. `ic-agent/src/agent/mod.rs`
```rust
pub fn signed_query_inspect;
pub fn signed_update_inspect;
pub fn signed_request_status_inspect;
```
These three public functions can be used to inspect the signed CBOR bytes and guarantee data authentication.
If there is a mismatch between the fields from arguments and from CBOR bytes, an `AgentError::CallDataMismatch` will be returned.

4. `ic-agent/src/agent/mod.rs`
```rust
impl Agent {
    pub fn query_signed;
    pub fn update_signed;
    pub fn request_status_signed;
}
```
These three methods are for sending the signed CBOR bytes to replica. Recommend using inspect functions in point 3 before sending.

ericswanson-dfinity:
I think it's good, as you have implemented, to
> [require] the user to provide all fields of the request together with the bytes
because the interface enforces that the actual data is checked.

ic-agent/src/agent/mod.rs
```diff
@@ -408,18 +391,45 @@ impl Agent {
...
+        let request_id = to_request_id(&request)?;
+        let serialized_bytes = sign_request(&request, self.identity.clone())?;
+
+        self.call_endpoint(effective_canister_id, request_id, serialized_bytes)
+            .await
+    }
+
+    /// Send the signed update to the network. Will return a [`RequestId`].
+    /// The bytes will be checked if it is a valid update.
+    /// If want to inspect the fields of the query, use [`signed_update_inspect`] before call this method.
```
ericswanson-dfinity, line 405:
```suggestion
    /// The bytes will be checked to verify that it is a valid update.
    /// If you want to inspect the fields of the query, use [`signed_update_inspect`] before calling this method.
```

ic-agent/src/agent/mod.rs
```diff
@@ -579,6 +593,27 @@ impl Agent {
         lookup_request_status(cert, request_id)
     }
 
+    /// Send the signed request_status to the network. Will return [`RequestStatusResponse`].
+    /// The bytes will be checked if it is a valid request_status.
+    /// If want to inspect the fields of the request_status, use [`signed_request_status_inspect`] before call this method.
```
ericswanson-dfinity, line 598:
```suggestion
    /// The bytes will be checked to verify that it is a valid request_status.
    /// If you want to inspect the fields of the request_status, use [`signed_request_status_inspect`] before calling this method.
```

ericswanson-dfinity (approved):
Looks great!  I just left a couple grammar corrections in comments for public functions.

ic-agent/src/agent/agent_error.rs
```diff
@@ -108,6 +108,13 @@ pub enum AgentError {
 
     #[error("An error happened during communication with the replica: {0}")]
     TransportError(Box<dyn std::error::Error + Send + Sync>),
+
+    #[error("There is mismatch between CBOR encoded call and the arguments: field {field}, value in argument is {value_arg}, value in CBOR is {value_cbor}")]
```
p-shahi, line 112:
```suggestion
    #[error("There is a mismatch between the CBOR encoded call and the arguments: field {field}, value in argument is {value_arg}, value in CBOR is {value_cbor}")]
```

ref-tests/tests/ic-ref.rs
```diff
@@ -770,3 +770,124 @@ mod extras {
         })
     }
 }
+
```
p-shahi, line 773: The tests in this file is meant to more or less mirror this Spec tests from this interface spec file: https://github.com/dfinity/ic-hs/blob/master/src/IC/Test/Spec.hs
Please move them to https://github.com/dfinity/agent-rs/blob/main/ref-tests/tests/integration.rs

ic-agent/src/agent/mod.rs
```diff
@@ -604,6 +639,229 @@ impl Agent {
...
+                    value_cbor: format!("{:?}", arg_cbor),
+                });
+            }
+        }
+    }
+    Ok(())
+}
+
+/// Inspect the bytes to be sent as a update
+/// Return Ok only when the bytes can be deserialized as a update and all fields match with the arguments
```
p-shahi, line 763:
```suggestion
/// Inspect the bytes to be sent as an update
/// Return Ok only when the bytes can be deserialized as an update and all fields match with the arguments
```

ic-agent/src/agent/mod.rs
```diff
@@ -375,26 +325,59 @@ impl Agent {
...
+                } => Err(AgentError::ReplicaError {
+                    reject_code,
+                    reject_message,
+                }),
+            })
+    }
+
+    /// Send the signed query to the network. Will return a byte vector.
+    /// The bytes will be checked if it is a valid query.
+    /// If want to inspect the fields of the query, use [`signed_query_inspect`] before call this method.
```
p-shahi, line 346:
```suggestion
    /// If you want to inspect the fields of the query call, use [`signed_query_inspect`] before calling this method.
```

ic-agent/src/agent/mod.rs
p-shahi, line 405:
```suggestion
    /// If want to inspect the fields of the update, use [`signed_update_inspect`] before call this method.
```

ic-agent/src/agent/signed.rs
```diff
@@ -0,0 +1,50 @@
...
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct SignedQuery {
+    pub ingress_expiry: u64,
+    pub sender: Principal,
+    pub canister_id: Principal,
+    pub method_name: String,
+    #[serde(with = "serde_bytes")]
+    pub arg: Vec<u8>,
+
```
p-shahi, line 14:
```suggestion
```

ic-agent/src/agent/signed.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub effective_canister_id: Principal,
+
```
p-shahi, line 16:
```suggestion
```

ic-agent/src/agent/signed.rs
```diff
@@ -0,0 +1,50 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(with = "serde_bytes")]
+    pub nonce: Option<Vec<u8>>,
+    pub ingress_expiry: u64,
+    pub sender: Principal,
+    pub canister_id: Principal,
+    pub method_name: String,
+    #[serde(with = "serde_bytes")]
+    pub arg: Vec<u8>,
+
```
p-shahi, line 33:
```suggestion
```

ic-agent/src/agent/signed.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub effective_canister_id: Principal,
+
```
p-shahi, line 35:
```suggestion
```

ic-agent/src/agent/signed.rs
```diff
@@ -0,0 +1,50 @@
...
+    #[serde(with = "serde_bytes")]
+    pub signed_update: Vec<u8>,
+
```
p-shahi, line 38:
```suggestion
```

p-shahi (approved):
awesome! just a few grammar comments

PR #204, opened 2021-06-09
fix: replica_api deserialization

Avi-D-coder:
`#[serde(skip_serializing_if = "Option::is_none")]` without `#[serde(default)]` causes deserialization to fail.


hansl (approved):
It does? Today I learnt... Thanks!

Avi-D-coder:
> It does? Today I learnt... Thanks!

https://serde.rs/attr-skip-serializing.html
> NOTE: Using skip_serializing does not skip deserializing the field. If you only add the skip_serializing attribute, and then attempt to deserialize the data, it will fail, as it will still attempt to deserialize the skipped field. Please use the skip attribute to skip both serializing and deserializing (see Field Attributes: skip). Likewise, use skip_deserializing to skip deserializing only.

Issue #208, opened 2021-06-10
Figure out why the Cargo.toml patch section does not work properly for candid and ic-types


hansl:
Upgrading the minor version of ic-types break the build.

Kinrany:
This also seemingly causes `ic-types` to break when referenced as a git dependency.

PR #211, opened 2021-06-14
fix: enable serde feature "derive" for non-dev

Kinrany:
This fixes at least the following case:

1. Clone agent-rs into `~/agent-rs`
2. Create a new project `~/ic-types-test`
3. Add a dependency on `../agent-rs/ic-types`
4. `~/ic-types-test` fails to build
5. Apply the change from this PR in `~/agent-rs`
6. `~/ic-types-test` builds

Fixes https://github.com/dfinity/agent-rs/issues/208


p-shahi (approved):

p-shahi:
Thanks for the patch!

PR #217, opened 2021-06-29
fix: fix CanisterStatus field rename

chenyan-dfinity:


ericswanson-dfinity (approved):

chenyan-dfinity:
CI still fails. Looking into it.

PR #218, opened 2021-06-30
chore: remove `ic-types`

Daniel-Bloom-dfinity:
Also update `candid` to `0.7.1`


icx-proxy/Cargo.toml
```diff
@@ -18,14 +18,13 @@ path = "src/main.rs"
 
 [dependencies]
 anyhow = "1.0.34"
-candid = "0.7"
+candid = "0.7.1"
```
krpeacock, line 21: Does leaving it with the minor version allow cargo to automatically adopt future patch updates, or was this necessary?

icx/Cargo.toml
```diff
@@ -17,13 +17,12 @@ name = "icx"
 path = "src/main.rs"
 
 [dependencies]
-candid = "0.7"
+candid = "0.7.1"
```
krpeacock, line 20: same question here

ref-tests/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-candid = "0.7"
+candid = "0.7.1"
```
krpeacock, line 9: and here

icx-proxy/Cargo.toml
ericswanson-dfinity: It does, but since this is for an executable, including the minor version is appropriate.

ic-utils/Cargo.toml
```diff
@@ -16,10 +16,9 @@ include = ["src", "Cargo.toml", "../LICENSE", "README.md"]
 
 [dependencies]
 async-trait = "0.1.40"
-candid = "0.7"
+candid = "0.7.1"
```
ericswanson-dfinity, line 19: I believe excluding the minor version here is intended, since this is a library.

ref-tests/Cargo.toml
ericswanson-dfinity: Since the rest here seem to specify minor version too, I think it's ok

icx-proxy/Cargo.toml
Daniel-Bloom-dfinity: We need this to ensure `candid` version `0.7.1` or higher in order to ensure the new `ic-types` is used. Yes, a cargo update would automagically use `0.7.1`, but this change is to force the semantic requirement that we need this patch version or greater (i.e. to prevent downstream packages from even being able to shoot themselves in the foot with `candid = "=0.7.0"`).

icx/Cargo.toml
Daniel-Bloom-dfinity: Same as above.

ref-tests/Cargo.toml
Daniel-Bloom-dfinity: Same as above.

ic-utils/Cargo.toml
Daniel-Bloom-dfinity: Same as above.

ericswanson-dfinity (approved):

Issue #203, opened 2021-06-05
Unable to depend on `ic-utils` for `http_request` interface; use of `IDLValue` in `HttpResponse` causes `candid` to panic


paulyoung:
As mentioned in https://github.com/dfinity/agent-rs/issues/201, I was hoping to depend on the `http_request` interface from `ic-utils`.

Having worked around that issue I ran into another: `Panicked at 'internal error: entered unreachable code'` in the `candid` value parser.

@chenyan-dfinity had some thoughts on how to address this in https://github.com/dfinity/candid/issues/242

PR #221, opened 2021-07-02
fix: icx-proxy panic when decoding a response to http_request

ericswanson-dfinity:
Candid 0.7.1 does not support decoding a Record into an IDLValue.  We were using this in an unsupported way with Candid 0.6.x, but it doesn't work anymore.

So, this commit specifies the exact type of the streaming http response token.  Unfortunately, this means that we will no longer support arbitrary token types, but this is the best that we can do until Candid supports generic types: https://github.com/dfinity/candid/issues/245


paulyoung:
Does this fix https://github.com/dfinity/agent-rs/issues/203?

ericswanson-dfinity:
> Does this fix #203?

Yes, I think it will.  We were running into errors like ```thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code', /sources/candid-0.7.1/src/types/subtype.rs:116:25```

chenyan-dfinity (approved):
LGTM

PR #223, opened 2021-07-03
chore: Replace ring by ed25519-dalek

Sawchord:
This PR replaces the ring library used inside of the identity module with ed25519-dalek.

The reason for this is that ring has some portability issues, that can be avoided this way.
Furthermore, ed25519-dalek is also already used within ic code.

Since ed25519-dalek does not come with a pkcs8 library, I added one in.
This requires some manual parsing.
Since this library uses const generics, the Rust version was raised to 1.52.1.

Please wait with merging this PR until some further reviews have been conducted.


ic-agent/src/identity/basic.rs
```diff
@@ -28,15 +30,56 @@ impl BasicIdentity {
...
+        }
+
+        // Check that there are no parameters on the key
+        if key_info.algorithm.parameters.is_some() {
+            return Err(PemError::KeyRejected(
+                "Parameters on the key are not allowed".to_string(),
+            ));
+        }
+
+        // Retrieve the secret key and check that we have an uncompressed
```
andreacerulli, line 52: Note that the secret key is not an `uncompressed point` but it is a 32 byte string.

ic-agent/src/identity/basic.rs
```diff
@@ -28,15 +30,56 @@ impl BasicIdentity {
...
+        // point of 32 byte length
+        let sk = key_info.private_key;
+        if sk[0] != 4 || sk[1] != 32 {
```
andreacerulli, line 55: If I understand correctly this only checks that the second byte of the encoding is equal to 32. Should we also check that `sk[2..]` has indeed length 32? Or is this already guaranteed/implicitly checked elsewhere?

Note that even if the stored key is longer the additional bytes will be truncated in the code below.

ic-agent/src/identity/basic.rs
```diff
@@ -28,15 +30,56 @@ impl BasicIdentity {
...
+            return Err(PemError::KeyRejected(
+                "Key is not a ECC private key".to_string(),
+            ));
+        }
+
+        // Retrieve the public key, error if it was not provided
+        let pk = match key_info.public_key {
+            Some(pk) => pk,
```
andreacerulli, line 63: Should we also check the length of the public key?

andreacerulli (commented):
Thanks @Sawchord! I left some minor comments below. I think it would be good to add some small tests, but other than that it looks good.

ic-agent/src/identity/basic.rs
Sawchord: IMO we shouldn't require the PK at all :( 
Instead of checking the public key, I will regerate the `PublicKey` from the `SecretKey` and then compare the two and reject, if they don't match.

ic-agent/src/identity/basic.rs
Sawchord: Done

ic-agent/src/identity/basic.rs
Sawchord: Will check that the sk marray is exact 34 bytes on total

ic-agent/src/identity/basic.rs
```diff
@@ -28,15 +30,63 @@ impl BasicIdentity {
         let bytes: Vec<u8> = pem_reader
             .bytes()
             .collect::<Result<Vec<u8>, std::io::Error>>()?;
+        let file = std::str::from_utf8(&bytes).unwrap();
```
Daniel-Bloom-dfinity, line 33: Replace with `pem_reader.read_to_string(&mut file)`.

ic-agent/src/identity/basic.rs
```diff
@@ -28,15 +30,63 @@ impl BasicIdentity {
...
+        let pem_file: PrivateKeyDocument = file.parse().map_err(|_| PemError::PemError)?;
+        let key_info = pem_file.private_key_info();
+
+        // Check OID
+        if key_info.algorithm.oid != "1.3.101.112".parse().unwrap() {
```
Daniel-Bloom-dfinity, line 38: Use a constant `const OID: ObjectIdentifier = ObjectIdentifier::new("1.3.101.112")`

Daniel-Bloom-dfinity (approved):

PR #229, opened 2021-07-13
feat: Add an option to let icx read from stdin

ninegua:
Taking call arguments from command line would exceed max limit for big inputs. This adds support to read from stdin when the input is `-`.

Also fixed a minor problem to properly exit with non-zero exit code when error.


lwshang (approved):

ericswanson-dfinity:
Sorry, I didn't manage to get this merged before some other changes with conflicts were merged.

PR #232, opened 2021-07-15
feat: icx sign/send with new ic-agent API

lwshang:
The implementation has been manually tested with the example in icx README.

We still lack of a test framework for cli package.

### DIFF
--- a/icx/README.md
+++ b/icx/README.md
@@ -22,8 +22,8 @@ To read a PEM file, you can pass it with the `--pem` argument. The PEM file must
 key that can be used for the Internet Computer signing and validation.
 
 ### Root Key
-For non-IC networks, pass --fetch-root-key to fetch the root key.  When this argument is not present,
-icx uses the hardcoded public key.
+For non-IC networks, pass `--fetch-root-key` to fetch the root key.  When this argument is not present,
+icx uses the hard-coded public key.
 
 ### Examples
 To call the management canister's `create_canister` function, you can use the following:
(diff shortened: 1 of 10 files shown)


ericswanson-dfinity:
FYI, we now have some bats-based tests for the cli tools, for example: https://github.com/dfinity/agent-rs/blob/main/e2e/bash/icx-asset.bash#L23

icx/src/main.rs
```diff
@@ -511,62 +455,77 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-                    println!("{}", hex::encode(result));
+            println!("{}", buffer);
+
+            let agent = Agent::builder()
+                .with_transport(
+                    agent::http_transport::ReqwestHttpReplicaV2Transport::create(
+                        opts.replica.clone(),
+                    )?,
+                )
+                .build()?;
```
ericswanson-dfinity, line 475: Unless I'm mistaken, here we've created two agents: one with a random key, and this one.  Could we make it so we only create the agent that we're going to use, in both cases?

icx/README.md
```diff
@@ -38,3 +38,31 @@ If you have a candid file, you can use it to validate arguments. Pass it in with
...
+
+#### Sign and Send Separation
+Pass `--serialize` when use `icx update` and `icx query` will serialize the signed canister call message as json.
+The output will print to stdout. When use `icx update`, a corresponding request_status message is also generated and printed to stderr.
+
+In the default IC project generated with `dfx new` and the local emulator has started with `dfx start --background`.
+
+##### Sign
+```shell script
+icx --fetch-root-key update --serialize rwlgt-iiaaa-aaaaa-aaaaa-cai greet '("everyone")' 1> update.json 2>request_status.json
```
ninegua, line 50: I will not recommend outputing useful data to standard error. For more than one output, we can print them as a JSON array instead.

icx/README.md
```diff
@@ -38,3 +38,31 @@ If you have a candid file, you can use it to validate arguments. Pass it in with
...
+##### Send
+```shell script
+cat update.json | icx send
+...
+RequestID: 0x1234....
+```
+
+##### Request status
+```shell script
+cat request_status.json | icx --fetch-root-key send
```
ninegua, line 63: If the default is hard-coded key, this `--fetch-root-key` doesn't seem necessary.

ericswanson-dfinity (approved):

PR #235, opened 2021-07-20
fix: run ic ref tests against pr branch

ericswanson-dfinity:
Prerequisite for merging this pr: removal of the IC_REF_TOKEN from github repo settings.

Since we're now using the public ic-hs repo, we don't need a github token in order to access it, and so we can run CI against the PR branch (`pull_request`) rather than the PR's base branch (`pull_request_target`).

### DIFF
--- a/.github/workflows/ic-ref.yml
+++ b/.github/workflows/ic-ref.yml
@@ -80,6 +80,7 @@ jobs:
       - name: Build wallet canister
         run: |
           cd wallet-rs/
+          echo "[workspace]" >>Cargo.toml
           sh ./wallet/build.sh
           cp wallet/target/wasm32-unknown-unknown/release/wallet-opt.wasm $HOME/wallet.wasm
 
--- a/.github/workflows/ic-ref.yml
+++ b/.github/workflows/ic-ref.yml
@@ -4,11 +4,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/ic-ref.yml
```diff
@@ -79,12 +73,14 @@ jobs:
       - name: Build universal-canister
         run: |
           cd ic-hs/universal-canister
+          echo "[workspace]" >>Cargo.toml
```
ericswanson-dfinity, line 76: This hack is to address the following error, which is appearing now:
```
error: current package believes it's in a workspace when it's not:
current:   /home/runner/work/agent-rs/agent-rs/ic-hs/universal-canister/Cargo.toml
workspace: /home/runner/work/agent-rs/agent-rs/Cargo.toml

this may be fixable by adding `ic-hs/universal-canister` to the `workspace.members` array of the manifest located at: /home/runner/work/agent-rs/agent-rs/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

This wasn't detected in the PR that updated to use ic-hs, because as noted in this PR, the ic-ref tests in that PR ran against the github workflow scripts on main.

krpeacock (approved):

PR #246, opened 2021-08-04
refactor: improve icx error handling with anyhow

lwshang:
### Benefit from this refactoring
1. Use `.context()` and `.with_context()` everywhere to get error stack
2. Running with `RUST_BACKTRACE=1` will print out the backtrace so that developer can easily locate the problem


icx/src/main.rs
```diff
@@ -139,49 +140,71 @@ struct PrincipalConvertOpts {
...
     method_name: &str,
-) -> Option<(TypeEnv, Function)> {
-    let (env, ty) = check_candid_file(idl_path).ok()?;
-    let actor = ty?;
-    let method = env.get_method(&actor, method_name).ok()?.clone();
-    Some((env, method))
+) -> Result<Option<(TypeEnv, Function)>> {
+    let (env, ty) = check_candid_file(idl_path).with_context(|| {
+        format!(
+            "Failed when check the candid file: {}",
```
ericswanson-dfinity, line 146: Grammar suggestion:
```suggestion
            "Failed when checking candid file: {}",
```

icx/src/main.rs
```diff
@@ -139,49 +140,71 @@ struct PrincipalConvertOpts {
...
-            let arguments = args.unwrap_or("()");
-            let args: Result<IDLArgs, String> = arguments
-                .parse::<IDLArgs>()
-                .map_err(|e: candid::Error| format!("Invalid Candid values: {}", e));
+            let arguments = arguments.unwrap_or("()");
+            let args = arguments.parse::<IDLArgs>();
             let typed_args = match method_type {
-                None => args?.to_bytes(),
+                None => args
+                    .context("Failed to parse arguments as Candid values")?
```
ericswanson-dfinity, line 206: Can we have this context message and the one below use different text, so that we can differentiate the cases?

icx/src/main.rs
```diff
@@ -270,8 +291,8 @@ pub fn get_effective_canister_id(
                 struct In {
                     canister_id: Principal,
                 }
-                let in_args = candid::Decode!(arg_value, In)
-                    .map_err(|err| format!("Candid decode err: {}", err))?;
+                let in_args =
+                    candid::Decode!(arg_value, In).context("Argument is not a valid Principle")?;
```
ericswanson-dfinity, line 295:
```suggestion
                    candid::Decode!(arg_value, In).context("Argument is not a valid Principal")?;
```

icx/src/main.rs
```diff
@@ -387,28 +418,35 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
                     }
                     Err(AgentError::TransportError(_)) => return Ok(()),
                     Err(AgentError::HttpError(HttpErrorPayload {
                         status,
                         content_type,
                         content,
                     })) => {
-                        eprintln!("Server returned an HTTP Error:\n  Code: {}", status);
+                        let mut error_message =
+                            format!("Server returned an HTTP Error:\n  Code: {}", status);
```
ericswanson-dfinity, line 430: Suggestion: add newlines at the end of these, like `eprintln` would have added, here and below:
```suggestion
                        let mut error_message =
                            format!("Server returned an HTTP Error:\n  Code: {}\n", status);
```

ericswanson-dfinity (changes requested):

icx/src/main.rs
lwshang: Sure, will do

ericswanson-dfinity (approved):

PR #248, opened 2021-08-09
feat!: icx: remove fetch-root-key parameter; determine automatically.

ericswanson-dfinity:
Fetch root key if replica address is not `https://ic0.app`

Breaking change: removes `--fetch-root-key` parameter

### DIFF
--- a/icx/src/main.rs
+++ b/icx/src/main.rs
@@ -26,6 +26,8 @@ use std::{
     str::FromStr,
 };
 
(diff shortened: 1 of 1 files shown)


jwiegley (approved):

nomeata:
Maybe not so important for `icx`, but a warning in general: such behaviour opens an attack vector where someone tricks someone else into using `https://ic0.app:443` or `https://identity.ic0.app`  – all valid “endpoints” that look the same to the uninitiated, but silently change behavior in a dangerous way.

PR #255, opened 2021-08-20
feat: expose `NonceGenerator` field via methods

Daniel-Bloom-dfinity:
Replace `NonceFactory` with trait and generics to give more control.
Adds some renaming to keep backwards compatibility, but transient types like `QueryBuilder`, which are not expected to be explicitly named, retain their old names just with default generics, resulting technically breaking back compat (but not realisticially for any actual users).

### DIFF
--- a/ic-agent/src/agent/builder.rs
+++ b/ic-agent/src/agent/builder.rs
@@ -1,11 +1,13 @@
 use crate::{
-    agent::{AgentConfig, ReplicaV2Transport},
-    Agent, AgentError, Identity, NonceFactory,
+    agent::{agent_config::AgentConfigImpl, AgentImpl, ReplicaV2Transport},
+    AgentError, Identity, NonceFactory, NonceGenerator,
 };
 use std::sync::Arc;
 
-pub struct AgentBuilder {
-    config: AgentConfig,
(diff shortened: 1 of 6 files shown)


ic-agent/src/agent/builder.rs
```diff
@@ -34,19 +36,36 @@ impl AgentBuilder {
...
+    pub fn with_nonce_generator<N1: NonceGenerator>(
+        self,
+        nonce_factory: N1,
+    ) -> AgentBuilderImpl<N1> {
+        AgentBuilderImpl {
+            config: AgentConfigImpl {
+                nonce_factory,
+                identity: self.config.identity,
+                ingress_expiry_duration: self.config.ingress_expiry_duration,
+                transport: self.config.transport,
```
ericswanson-dfinity, line 68: Could this be like so?
```suggestion
                ...self.config
```

ericswanson-dfinity (approved):

ic-agent/src/agent/builder.rs
Daniel-Bloom-dfinity: I wish. But since they're not the same type (`N` is different), it gets all mad.

PR #271, opened 2021-09-24
Add an option to fetch the root key, for testing or local development.

jplevyak:
Signed-off-by: John Plevyak <jplevyak@gmail.com>

### DIFF
--- a/icx-proxy/src/main.rs
+++ b/icx-proxy/src/main.rs
@@ -84,6 +84,11 @@ pub(crate) struct Opts {
     #[clap(long)]
     debug: bool,
 
+    /// Whether or not to fetch the root key from the replica back end. Do not use this when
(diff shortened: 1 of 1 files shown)


icx-proxy/src/main.rs
```diff
@@ -459,8 +472,11 @@ async fn handle_request(
                 .build()
                 .expect("Could not create agent..."),
         );
-
-        forward_request(request, agent, dns_canister_config.as_ref(), logger.clone()).await
+        if fetch_root_key && agent.fetch_root_key().await.is_err() {
```
ericswanson-dfinity, line 475: fetch_root_key is meant to be called once, after constructing the agent, rather than many times preceding individual operations.

paulyoung:
For what it’s worth, I did this in https://github.com/dfinity/agent-rs/pull/195

PR #280, opened 2021-11-02
feat: add url rewrite to `ReqwestHttpReplicaV2Transport`

Daniel-Bloom-dfinity:
Rewrite all `*.ic0.app` domains to `ic0.app` to avoid redirects.


ic-agent/src/agent/http_transport.rs
```diff
@@ -43,7 +45,15 @@ impl ReqwestHttpReplicaV2Transport {
 
         Ok(Self {
             url: reqwest::Url::parse(&url)
-                .and_then(|url| url.join("api/v2/"))
+                .and_then(|mut url| {
+                    // rewrite *.ic0.app to ic0.app
+                    if let Some(domain) = url.domain() {
+                        if domain.ends_with(IC0_DOMAIN) && domain != IC0_DOMAIN {
```
ericswanson-dfinity, line 51: any concerns about case sensitivity?

Daniel-Bloom-dfinity: Thankfully `domain`s are all encoded as lowercase as per IDNA spec, so case sensitivity is not a problem. I've added some tests to cover this just to make doubly sure.

ic-agent/src/agent/http_transport.rs
```diff
@@ -263,10 +263,11 @@ mod test {
...
 
         test("https://ic0.app", "https://ic0.app/api/v2/");
+        test("https://IC0.app", "https://ic0.app/api/v2/");
         test("https://foo.ic0.app", "https://ic0.app/api/v2/");
-        test("https://foo.bar.ic0.app", "https://ic0.app/api/v2/");
-        test("https://ic0.app/foo/", "https://ic0.app/foo/api/v2/");
-        test("https://foo.ic0.app/foo/", "https://ic0.app/foo/api/v2/");
+        test("https://foo.IC0.app", "https://ic0.app/api/v2/");
+        test("https://foo.Ic0.app", "https://ic0.app/api/v2/");
+        test("https://foo.iC0.app", "https://ic0.app/api/v2/");
```
ericswanson-dfinity, line 270: Could you add rather than replace these tests?  Two of the replaced tests had output that differed from the others (now every test output ends in `/ic?.app/api/v2/`)

Daniel-Bloom-dfinity: Whoops. Meant to just add test lines. Copy-paste error.

ic-agent/src/agent/http_transport.rs
```diff
@@ -240,3 +250,30 @@ impl super::ReplicaV2Transport for ReqwestHttpReplicaV2Transport {
...
+    #[test]
+    fn redirect() {
+        fn test(base: &str, result: &str) {
+            let t = ReqwestHttpReplicaV2Transport::create(base).unwrap();
+            assert_eq!(t.url.as_str(), result, "{}", base);
+        }
+
+        test("https://ic0.app", "https://ic0.app/api/v2/");
+        test("https://IC0.app", "https://ic0.app/api/v2/");
+        test("https://foo.ic0.app", "https://ic0.app/api/v2/");
```
ericswanson-dfinity, line 267: What would be the expected behavior?
test("https://fooic0.app", "https://???/api/v2/");

Daniel-Bloom-dfinity: Good catch. Should be unchanged. Fixed and tested in next revision.

ericswanson-dfinity (approved):

PR #291, opened 2021-12-22
feat: Provide support for upgrading HTTP requests

paulyoung:
This lays the groundwork for upgrading HTTP requests from query calls to update calls.

It applies the changes from https://github.com/dfinity/agent-rs/pull/195 that are specific to `ic-utils` so that they can be used downstream in `icx-proxy`.


paulyoung:
This is the downstream PR: https://github.com/dfinity/icx-proxy/pull/6

ericswanson-dfinity (approved):

PR #295, opened 2021-12-31
feat: use generics for `lookup_value`

Daniel-Bloom-dfinity:
Replaces `Vec` with array for calls
Ready to switch to `Iterator`s for dfinity/ic-types#8

### DIFF
--- a/ic-agent/src/agent/response_authentication.rs
+++ b/ic-agent/src/agent/response_authentication.rs
@@ -48,7 +48,7 @@ pub(crate) fn lookup_canister_info(
     canister_id: Principal,
     path: &str,
 ) -> Result<Vec<u8>, AgentError> {
-    let path_canister = vec!["canister".into(), canister_id.into(), path.into()];
+    let path_canister = ["canister".into(), canister_id.into(), path.into()];
     lookup_value(&certificate, path_canister).map(<[u8]>::to_vec)
 }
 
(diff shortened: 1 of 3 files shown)


ic-agent/src/agent/response_authentication.rs
```diff
@@ -147,14 +145,19 @@ pub(crate) fn lookup_reply(
     Ok(RequestStatusResponse::Replied { reply })
 }
 
-pub fn lookup_value<'a>(
+pub fn lookup_value<'a, P>(
```
ericswanson-dfinity, line 148: Would you please add an entry to the changelog since it looks like technically this is a breaking change?

ericswanson-dfinity, line 148: Would you please add an entry to the changelog since it looks like technically this is a breaking change?

ericswanson-dfinity (changes requested):

ic-agent/src/agent/response_authentication.rs
Daniel-Bloom-dfinity: Actually this is not a breaking change, since `Vec<Label>` implements the required interfaces, but I've added to the changelog anyway.

ericswanson-dfinity (approved):

ic-agent/src/agent/response_authentication.rs
ericswanson-dfinity: Even better! Thanks!

PR #296, opened 2022-01-07
chore: release rust agent 0.11.0

ericswanson-dfinity:
Updated candid and ic-types, along with minor updates to other dependencies.


adamspofford-dfinity:
This is a breaking change for the ic_types version bump, as anyone on 0.2.2 won't be able to use the same Principal type and updating to 3 breaks any code that was relying on the AsRef<[u8]> behavior.

jwiegley (approved):

PR #312, opened 2022-02-23
chore: add canister_range check

sesi200:
Jira: [FOLLOW-470](https://dfinity.atlassian.net/browse/FOLLOW-470)

### DIFF
--- a/ic-agent/src/agent/agent_test.rs
+++ b/ic-agent/src/agent/agent_test.rs
@@ -367,4 +367,5 @@ fn check_subnet_range_with_unauthorized_range() -> Result<(), AgentError> {
             .await
     });
     assert_eq!(result, Err(AgentError::CertificateNotAuthorized()));
+    Ok(())
 }
--- a/ic-agent/src/agent/agent_test.rs
+++ b/ic-agent/src/agent/agent_test.rs
@@ -8,6 +8,7 @@ use crate::{
         Status,
     },
     export::Principal,
+    hash_tree::Label,
     Agent, AgentError,
 };
 use mockito::mock;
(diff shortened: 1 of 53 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -618,13 +618,29 @@ impl Agent {
...
                     .map_err(AgentError::InvalidCborData)?;
-                self.verify(&cert)?;
+                self.verify(&cert, canister)?;
+                let canister_range_lookup = [
+                    "subnet".into(),
+                    delegation.subnet_id.clone().into(),
+                    "canister_ranges".into(),
+                ];
+                let canister_range = lookup_value(&cert, canister_range_lookup)?;
+                let ranges: Vec<Vec<Principal>> =
```
roman-kashitsyn, line 638:
```suggestion
                let ranges: Vec<(Principal, Principal)> =
```
This way we won't crash the agent if the canister ranges are invalid (contain a single principal instead of a range).

ic-agent/src/agent/agent_test.rs
```diff
@@ -236,3 +237,138 @@ fn status_error() -> Result<(), AgentError> {
...
+        Err(other) => panic!("found something unexpected: {:?}", other),
+    }
+}
+
+#[test]
+// asserts that a delegated certificate with /subnet/<subnetid>/canister_ranges that don't include
+// the canister gets rejected by the cert verification because the subnet is not authorized to
+// respond to requests for this canister. We do this by using a correct response but serving it
+// for the wrong canister, which a malicious node might do.
+fn check_subnet_range_with_unauthorized_range() -> Result<(), AgentError> {
```
nomeata, line 343: How about a test that checks that verification fails if `/subnet/<subnetid>/canister_ranges` is missing? (I think the code _does_ fail, but still)

ic-agent/src/agent/agent_test.rs
```diff
@@ -236,3 +237,138 @@ fn status_error() -> Result<(), AgentError> {
...
+                    .collect()],
+                wrong_canister,
+            )
+            .await
+    });
+    match result {
+        Err(AgentError::CertificateNotAuthorized()) => Ok(()),
+        Ok(_) => panic!("cert with bad canister_range considered valid"),
+        Err(other) => panic!("found something unexpected: {:?}", other),
+    }
```
roman-kashitsyn, line 373: nit: we could simplify this a bit:
```suggestion
    assert_eq!(result, Err(AgentError::CertificateNotAuthorized()));
```
I believe the error message in case of a test failure will be clear enough.

ic-agent/src/agent/mod.rs
sesi200: implemented

ic-agent/src/agent/agent_test.rs
```diff
@@ -236,3 +237,135 @@ fn status_error() -> Result<(), AgentError> {
...
+    228, 237, 162, 234, 22, 105, 115, 105, 103, 110, 97, 116, 117, 114, 101, 88, 48, 172, 233, 252,
+    221, 155, 201, 119, 224, 93, 99, 40, 248, 137, 220, 78, 124, 153, 17, 76, 115, 122, 73, 70, 83,
+    203, 39, 161, 245, 92, 6, 244, 85, 94, 15, 22, 9, 128, 175, 94, 173, 9, 138, 204, 25, 80, 16,
+    178, 247,
+];
+
+#[test]
+// asserts that a delegated certificate with correct /subnet/<subnetid>/canister_ranges
+// passes the certificate verification
+fn check_subnet_range_with_valid_range() -> Result<(), AgentError> {
```
roman-kashitsyn, line 307: Since you liked my nits: we don't need to return a result from a test, panics are preferred because they point you to a specific line where the error occurred. This makes debugging tests much easier.

roman-kashitsyn (approved):

ic-agent/src/agent/agent_test.rs
sesi200: was a huge pain, but I managed to create a response to add this test

sesi200:
Will pass CI once #317 is merged

ic-agent/src/agent/mod.rs
```diff
@@ -258,6 +258,14 @@ pub struct Agent {
     transport: Arc<dyn ReplicaV2Transport>,
 }
 
+impl fmt::Debug for Agent {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        f.debug_struct("Agent")
+            .field("ingress_expiry_duration", &self.ingress_expiry_duration)
+            .finish()
```
adamspofford-dfinity, line 265: If certain fields are absent, this should be notated in the output via `finish_non_exhaustive`.

sesi200: Good point, thanks!

lwshang (approved):

PR #316, opened 2022-03-03
chore: cycles are now u128 instead of u64

sesi200:
As requested on Slack


ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -48,7 +48,7 @@ pub struct CreateCanisterBuilder<'agent, 'canister: 'agent, T> {
     memory_allocation: Option<Result<MemoryAllocation, AgentError>>,
     freezing_threshold: Option<Result<FreezingThreshold, AgentError>>,
     is_provisional_create: bool,
-    amount: Option<u64>,
+    amount: Option<u128>,
```
ericswanson-dfinity, line 51: Please include a note in the [changelog](https://github.com/dfinity/agent-rs/blob/main/CHANGELOG.md)

sesi200: Good catch, thank you!

ericswanson-dfinity (approved):

PR #318, opened 2022-03-07
feat: Re-genericize tokens

Daniel-Bloom-dfinity:
re: dfinity/agent-rs#221 dfinity/agent-rs#203

Turns out you can actually get around candid's limitation on `IDLValue`. It just takes some careful rejiggering.

First, use `Reserved` to match anything. This seems like a valid use of `Reserved`, and should basically always be supported.

Second, use `deserialize_ignored_any` to actually get all the content with our own visitor (really just a copy of `IDLValueVisitor`). This is a little bit of a hack, but it works.


Daniel-Bloom-dfinity:
Should resolve [this issue](https://forum.dfinity.org/t/http-request-streaming-callback-does-not-stream-chunks-anymore/11298/9) with streaming tokens.

chenyan-dfinity (commented):
Interesting idea!

Is there a way to avoid the code duplication of `IDLValueVisitor`? Future candid changes may cause inconsistent behaviors in icx-proxy. Does it work if we expose `IDLValueVisitor` from the candid crate?

Daniel-Bloom-dfinity:
> Interesting idea!
> 
> Is there a way to avoid the code duplication of `IDLValueVisitor`? Future candid changes may cause inconsistent behaviors in icx-proxy. Does it work if we expose `IDLValueVisitor` from the candid crate?

It should. I literally just copied the code of `IDLValueVisitor` and fixed the import paths, so an export should work just as well.

ericswanson-dfinity (approved):

chenyan-dfinity (approved):

PR #323, opened 2022-03-17
fix: workaround for dfinity/candid#273

Daniel-Bloom-dfinity:
With a little elbow grease we can even get around `Func`'s limitations.

Introduces a new type `HttpRequestStreamingCallback` which properly serializes/deserializes callbacks of the form:
`callback : (Token) -> (StreamingCallbackHttpResponse) query;`

We use `ArgToken` (which returns `Type::Empty`  from `ty`) for the abstract form, because arguments subtype in the opposite direction so:

`callback : (Empty) -> (StreamingCallbackHttpResponse<Reserved>) query;`

Will match without knowing what the `Token` contents are.

### DIFF
--- a/ic-utils/src/interfaces/http_request.rs
+++ b/ic-utils/src/interfaces/http_request.rs
@@ -168,7 +168,7 @@ impl CandidType for ArgToken {
         Type::Empty
     }
     fn idl_serialize<S: Serializer>(&self, _serializer: S) -> Result<(), S::Error> {
-        // We cannot implement serialize, since our type must be `Reserved` in order to accept anything.
+        // We cannot implement serialize, since our type must be `Empty` in order to accept anything.
         // Attempting to serialize this type is always an error and should be regarded as a compile time error.
         unimplemented!("Token is not serializable")
     }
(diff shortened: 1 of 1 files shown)


ic-utils/src/interfaces/http_request.rs
```diff
@@ -101,6 +159,21 @@ impl<'de> Deserialize<'de> for Token {
...
+/// A marker type to match unconstrained callback arguments
+#[derive(Debug, Clone, Copy, PartialEq, Deserialize)]
+pub struct ArgToken;
+
+impl CandidType for ArgToken {
+    fn _ty() -> Type {
+        Type::Empty
+    }
+    fn idl_serialize<S: Serializer>(&self, _serializer: S) -> Result<(), S::Error> {
+        // We cannot implement serialize, since our type must be `Reserved` in order to accept anything.
```
chenyan-dfinity, line 171:
```suggestion
        // We cannot implement serialize, since our type must be `Empty` in order to accept anything.
```

chenyan-dfinity (approved):
LGTM as long as your tests are happy.

PR #327, opened 2022-03-29
chore: Update Rust to 1.58.1

adamspofford-dfinity:
This PR updates the rust-toolchain.toml version to 1.58.1, enables 2021 edition for all the packages, and adds the new MSRV key to Cargo.toml.


lwshang (approved):

ic-agent/src/identity/secp256k1.rs
```diff
@@ -23,7 +23,7 @@ use simple_asn1::{
 #[derive(Clone, Debug)]
 pub struct Secp256k1Identity {
     private_key: EcKey<Private>,
-    public_key: EcKey<Public>,
+    _public_key: EcKey<Public>,
```
ericswanson-dfinity, line 26: Is this now required to change to _public_key because it's only referenced in tests?

.github/workflows/release.yml
```diff
@@ -16,12 +16,12 @@ jobs:
             name: linux
             binary_path: target/x86_64-unknown-linux-musl/release
             binary_files: icx
-            rust: 1.50.0 # gmiam/rust-musl-action@master is only at 1.50.0 now
+            rust: '1.58.1'
```
ericswanson-dfinity, line 19: Would you please test this workflow before merging?  To do so, you'd add the branch name from this PR to `push: branches: ` above.

ic-agent/src/identity/secp256k1.rs
adamspofford-dfinity: Yes, that or #[allow(dead_code)]. In between the versions they removed the exemption for cfg(test) from the normal rules about usedness and cfg.

.github/workflows/release.yml
adamspofford-dfinity: Works great

ericswanson-dfinity (approved):

.github/workflows/release.yml
ericswanson-dfinity: Thanks!

adamspofford-dfinity:
Are those expected checks hardcoded to 1.55.0? Should I add aggregate checks to ease upgrading in the future?

ericswanson-dfinity:
> Are those expected checks hardcoded to 1.55.0? Should I add aggregate checks to ease upgrading in the future?

Yes, they are hardcoded unfortunately.  And yes, please go ahead and aggregate the checks, and I will update the required statuses.

.github/workflows/icx_asset.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ macos-latest, ubuntu-latest ]
-        rust: [ '1.55.0' ]
+        rust: [ '1.58.1' ]
```
ericswanson-dfinity, line 13: This one will need an aggregate as well

PR #328, opened 2022-03-29
feat: add ic-wasm to perform Wasm transformations for canister

chenyan-dfinity:
Part of https://dfinity.atlassian.net/browse/IC-1147

The library can be used to manage canister metadata. It can be called by dfx or as a standalone CLI tool. 

In the future, this library will also support other Wasm related features, such as canister profiling, code size reduction and inspection of IC system API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -914,6 +914,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "ic-wasm"
+version = "0.1.0"
+dependencies = [
+ "walrus",
+]
+
+[[package]]
 name = "icx"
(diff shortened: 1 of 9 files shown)


ericswanson-dfinity (changes requested):
Can we have a github workflow to build these?

chenyan-dfinity:
Moved to https://github.com/dfinity/ic-wasm

paulyoung:
Is https://github.com/dfinity/ic-wasm private?

chenyan-dfinity:
Will be public soon.

PR #334, opened 2022-04-12
fix: Fix clap deprecation

adamspofford-dfinity:
Fixes the new `clap` warning and updates `flate2`.


ericswanson-dfinity (approved):
Thanks!

PR #361, opened 2022-07-12
feat(ic-asset): configurable dot-directories and -files inclusion

smallstepman:
# Description
Closes https://dfinity.atlassian.net/browse/SDK-592.

#### include all hidden files
```json
[
    { "match": ".*", "ignore_hidden": false }
]
```
#### include all non-dotfiles from `.well-known` directory
```json
[
    { "match": ".well-known", "ignore_hidden": false }
]
```
#### include all non-dotfiles from `.well-known` directory + a specific dotfile from that dir
`/.ic-assets.json`
```json
[
    { "match": ".well-known", "ignore_hidden": false }
]
```
`/.well-known/.ic-assets.json`
```json
[
    { "match": ".dotfile", "ignore_hidden": false }
]
```
# How Has This Been Tested?

Unit tested `sync::gather_asset_descriptors`, see: https://github.com/dfinity/agent-rs/blob/c77fe91bb1559c1398e885ebf2bbc4052768303d/ic-asset/src/sync.rs#L142

# Checklist:


ic-asset/src/asset_config.rs
```diff
@@ -27,6 +27,7 @@ struct AssetConfigRule {
     r#match: GlobMatcher,
     cache: Option<CacheConfig>,
     headers: Maybe<HeadersConfig>,
+    ignore_hidden: bool,
```
ericswanson-dfinity, line 30: I'd suggest to make this about "ignoring files", rather than about "ignoring hidden files."

Where the default heuristic is:
- filename starts with a dot: ignore the file
- filename does not start with a dot: include the file

The `ignore` field, if present, overrides the above behavior (for any file)

```suggestion
    ignore: Option<bool>,
```

ic-asset/src/sync.rs
```diff
@@ -70,9 +79,13 @@ fn gather_asset_descriptors(dirs: &[&Path]) -> anyhow::Result<Vec<AssetDescripto
         let mut asset_descriptors_interim = vec![];
         for e in WalkDir::new(&dir)
             .into_iter()
-            .filter_entry(|entry| !filename_starts_with_dot(entry))
+            .filter_entry(|entry| {
+                doesnt_start_with_a_dot_or_explicitly_included(entry, &configuration)
+            })
```
ericswanson-dfinity, line 84:
```suggestion
            .filter_entry(|entry|include_entry(entry, &configuration))
```

ic-asset/src/sync.rs
```diff
@@ -49,12 +49,21 @@ pub async fn sync(
     Ok(())
 }
 
-fn filename_starts_with_dot(entry: &walkdir::DirEntry) -> bool {
-    entry
+fn doesnt_start_with_a_dot_or_explicitly_included(
+    entry: &walkdir::DirEntry,
+    config: &AssetSourceDirectoryConfiguration,
+) -> bool {
+    let starts_with_a_dot = entry
```
ericswanson-dfinity, line 56: Suggest reversing the meaning of this method (return true to include the entry)
```suggestion
fn include_entry(
    entry: &walkdir::DirEntry,
    config: &AssetSourceDirectoryConfiguration,
) -> bool {
    let starts_with_a_dot = entry
```

ic-asset/src/asset_config.rs
```diff
@@ -214,6 +238,18 @@ impl AssetConfigTreeNode {
...
+        let included_in_parent = match &self.parent {
+            Some(parent) => parent.is_included(canonical_path),
+            None => false,
+        };
+        self.rules
+            .iter()
+            .filter(|rule| rule.applies(canonical_path))
+            .any(|rule| !rule.ignore_hidden)
+            || included_in_parent
+    }
```
ericswanson-dfinity, line 252: I think we can simplify this by removing the "hidden" filtering from the directory walk, and instead merging the `ignore` field into the AssetConfig and filtering using that.

ic-asset/src/sync.rs
smallstepman: change commited, but still would like to learn more about the idea

smallstepman: I don't think removing `filter_entry` is a good idea, because there are cases where it gives a nice performance boost:
https://docs.rs/walkdir/latest/walkdir/struct.IntoIter.html#method.filter_entry
```
pub fn filter_entry(...) -> bool 
 
Yields only entries that satisfy the given predicate and skips descending into directories
that do not satisfy the given predicate.
The predicate is applied to all entries. If the predicate is true, iteration carries on as normal. 
If the predicate is false, the entry is ignored and if it is a directory, it is not descended into.
```

So if somebody would create a `.git` repo inside the assets dir, the directory would be skipped entirely thanks to `filter_entry`. 

Sadly, the `AssetSourceDirectoryConfiguration::config_map` would still get populated with all directories from `.git`, but that's because we went with `std::fs::read_dir(&dir)` instead of `walkdir`, therefore there is no way to take advantage of `filter_entry` there. We might want to change that in the future.

ic-asset/src/asset_config.rs
smallstepman: i'll explore this idea tomorrow

ic-asset/src/asset_config.rs
smallstepman: While I think this suggestion is a little bit outside of what this PR is trying to achieve (the PR was conceived due to the need to include `.well-known/*` assets in asset upload), I think it's overall a good idea, and makes everything "endgame-customizable".

I'll include this in this PR, because I don't want our users to get confused about the change 
> `ignore_hidden` is not `ignore`

ic-asset/src/sync.rs
smallstepman: e.g.
```rust
            .filter_entry(|entry| configuration.get_config(entry).is_included())
```

```rust
impl AssetSourceDirectoryConfiguration {
  fn get_config -> ConfigFileResult<AssetConfig> {...}
}

enum ConfigFileResult {
  NotFound,
  Excluded,
  Included<AssetConfig>
}

impl ConfigFileResult {
  is_included(&self) -> bool { 
    self == Self::Included 
} //pseudo
}
```
?

ic-asset/src/asset_config.rs
smallstepman: done in https://github.com/dfinity/agent-rs/pull/361/commits/ffc5abe70058880f9f92c5ebfb3f215944e834ed

turned out great!

ic-asset/src/sync.rs
smallstepman: done in https://github.com/dfinity/agent-rs/pull/361/commits/ffc5abe70058880f9f92c5ebfb3f215944e834ed

turned out great, thanks for sharing your idea!

smallstepman:
please advise on this Windows-specific error:
- error - https://github.com/dfinity/agent-rs/runs/7322654802?check_suite_focus=true
- source - https://github.com/dfinity/agent-rs/blob/SDK-592-configurable-hidden-directories-files-inclusion-when-uploading-assets/ic-asset/src/sync.rs#L372

do you think it's ok to put it behind `#[cfg(target_family = "unix")]`?

ic-asset/src/asset_config.rs
```diff
@@ -227,6 +232,10 @@ impl AssetConfig {
             (_, Maybe::Null) => self.headers = None,
             (_, Maybe::Absent) => (),
         };
+
+        if self.ignore != Some(true) && other.ignore.is_some() {
+            self.ignore = other.ignore;
+        }
```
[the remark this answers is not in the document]
smallstepman: Because it shouldn't be possible to unignore previously ignored files. 
I tried to follow the logic from https://git-scm.com/docs/gitignore
> An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again. It is not possible to re-include a file if a parent directory of that file is excluded. Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined. Put a backslash ("\") in front of the first "!" for patterns that begin with a literal "!", for example, "\!important!.txt".

accepting proposed change would fail the following test 
https://github.com/dfinity/agent-rs/blob/SDK-592-configurable-hidden-directories-files-inclusion-when-uploading-assets/ic-asset/src/sync.rs#L551
but changing 
```rust
let mut expected_asset_descriptors = vec![
    AssetDescriptor::default_from_path(&assets_dir, "file"),
    AssetDescriptor::default_from_path(&assets_dir, ".hidden-dir-flat/.hfile"),
];
``` 
to 
```rust
let mut expected_asset_descriptors = vec![
    AssetDescriptor::default_from_path(&assets_dir, "file"),
    AssetDescriptor::default_from_path(&assets_dir, ".hidden-dir-flat/.hfile"),
    AssetDescriptor::default_from_path(&assets_dir, ".hidden-dir-flat/file"),
];
```
would make it pass again.

Let me know how you'd want to proceed here.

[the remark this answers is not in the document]
ericswanson-dfinity: This is the exact case that I was wondering about.

I would expect the following:
1. If a directory is ignored, file and directories within it cannot be un-ignored
2. Properties from a rule matching a file in a subdirectory override properties from a rule matching a file in a parent directory

The proposed change achieves this, and is also consistent with the referenced rules for .gitnore:

> ... any matching file excluded by a previous pattern will become included again. It is not possible to re-include a file if a parent directory of that file is excluded.

What do you think?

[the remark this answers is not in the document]
smallstepman: sold!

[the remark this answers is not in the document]
smallstepman: done in https://github.com/dfinity/agent-rs/pull/361/commits/1c95f701ace6d9ed9c81fe810fbc60c814250f89

ic-asset/src/sync.rs
```diff
@@ -124,3 +140,564 @@ fn assemble_synchronization_operations(
...
+    }
+
+    impl PartialEq for AssetDescriptor {
+        fn eq(&self, other: &Self) -> bool {
+            [
+                self.source == other.source,
+                self.key == other.key,
+                self.config.cache == other.config.cache,
+                self.config.headers == other.config.headers,
+                self.config.ignore.map_or(false, |v| v) == other.config.ignore.map_or(false, |v| v),
```
ericswanson-dfinity, line 192: equivalent?
```suggestion
                self.config.ignore.unwrap_or(false) == other.config.ignore.unwrap_or(false),
```

ic-asset/src/sync.rs
```diff
@@ -124,3 +140,564 @@ fn assemble_synchronization_operations(
...
+            ),
+        ];
+
+        expected_asset_descriptors.sort_by_key(|v| v.source.clone());
+        asset_descriptors.sort_by_key(|v| v.source.clone());
+        assert_eq!(asset_descriptors, expected_asset_descriptors);
+    }
+
+    #[test]
+    fn impossible_to_reinclude_already_ignored_file() {
```
ericswanson-dfinity, line 551:
```suggestion
    fn possible_to_reinclude_previously_ignored_file() {
```

ericswanson-dfinity (approved):

PR #378, opened 2022-08-26
fix disable_range_check condition

mraszyk:
The `disable_range_check` condition should check that the call is to the Management Canister, not to an application canister offering a method with the name `provisional_canister_create_with_cycles`.

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -1192,10 +1192,10 @@ pub struct UpdateBuilder<'agent> {
 impl<'agent> UpdateBuilder<'agent> {
     /// Creates a new query builder with an agent for a particular canister method.
     pub fn new(agent: &'agent Agent, canister_id: Principal, method_name: String) -> Self {
-        // When calling provisional_create_canister_with_cycles, every effective_canister_id is valid.
+        // When calling provisional_create_canister_with_cycles on the Management Canister, every effective_canister_id is valid.
         // Therefore we need to disable the check for valid canister_ranges in the certificate validation.
         // More info: https://docs.dfinity.systems/spec/public/#http-effective-canister-id
-        let disable_range_check = method_name == "provisional_create_canister_with_cycles";
+        let disable_range_check = canister_id == Principal::from_text("aaaaa-aa").unwrap() && method_name == "provisional_create_canister_with_cycles";
         Self {
             agent,
             effective_canister_id: canister_id,


ic-agent/src/agent/mod.rs
```diff
@@ -1192,10 +1192,10 @@ pub struct UpdateBuilder<'agent> {
 impl<'agent> UpdateBuilder<'agent> {
     /// Creates a new query builder with an agent for a particular canister method.
     pub fn new(agent: &'agent Agent, canister_id: Principal, method_name: String) -> Self {
-        // When calling provisional_create_canister_with_cycles, every effective_canister_id is valid.
+        // When calling provisional_create_canister_with_cycles on the Management Canister, every effective_canister_id is valid.
         // Therefore we need to disable the check for valid canister_ranges in the certificate validation.
         // More info: https://docs.dfinity.systems/spec/public/#http-effective-canister-id
-        let disable_range_check = method_name == "provisional_create_canister_with_cycles";
+        let disable_range_check = canister_id == Principal::from_text("aaaaa-aa").unwrap() && method_name == "provisional_create_canister_with_cycles";
```
ericswanson-dfinity, line 1198:
```suggestion
        let disable_range_check = canister_id == Principal::management_canister() && method_name == "provisional_create_canister_with_cycles";
```

ericswanson-dfinity (changes requested):

ic-agent/src/agent/mod.rs
```diff
@@ -242,6 +243,11 @@ pub struct Agent {
     transport: Arc<dyn ReplicaV2Transport>,
 }
 
+#[derive(CandidType, Deserialize)]
+struct CreateCanisterResult {
+  canister_id: candid::Principal,
```
mraszyk, line 248: Why cannot I change `candid::Principal` to `Principal` here? I could change it if I replace `let paths = cert.tree.list_paths();` by `let paths: Vec<Vec<Label> > = vec![];` on l. 631--632 and drop `git = ...` for the `ic-types` dependency in `Cargo.toml`.

mraszyk: Resolved thanks to @Daniel-Bloom-dfinity

sesi200:
@mraszyk any chance you can document this code better somehow? I have a really hard time understanding what happens for what reason

mraszyk:
@sesi200 Absolutely, I'll add more comments into the code. Ultimately, the goal is to make the agent compliant with this [part](https://ic-interface-spec.netlify.app/#http-read-state) of the spec.

sesi200 (approved):

ic-agent/src/agent/mod.rs
```diff
@@ -633,6 +663,50 @@ impl Agent {
...
+        if disable_range_check && contains_delegations && !is_delegation {
+            let paths = cert.tree.list_paths();
+            let t: Label = "time".into();
+            let rs: Label = "request_status".into();
+            // rid contains <request_id> from paths /request_status/<request_id> if present in the certificate
+            let mut rid: Option<Label> = None;
+            for p in paths {
+                if !(p == vec![t.clone()] || p[0] == rs) {
+                    // the path has neither the form /time nor /request_status/*
+                    return Err(AgentError::CertificateVerificationFailed());
```
ericswanson-dfinity, line 676: Up until now, the `CertificateVerificationFailed` error referred to a specific cause: `bls::core_verify()` returned something other than `BLS_OK`, meaning the certificate signature didn't match the expected value.

I think it woiuld be helpful to someone diagnosing these errors if we introduced a new error type, `CertificateVerificationRejected(reason)`, where reason is an enumeration that gives more information about the reason for the rejection of the certificate.  What do you think?

ericswanson-dfinity (changes requested):

mraszyk:
Superseded by this [PR](https://github.com/dfinity/agent-rs/pull/389).

PR #389, opened 2022-10-14
feat: drop disable_range_check flag

mraszyk:
The pre-master and hourly tests of the IC pass upon this change.

TODO(MR): fix ic-ref tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.22.0] - 2022-10-14
+
+* Drop `disable_range_check` flag from certificate delegation checking.
+
 ## [0.21.0] - 2022-10-03
 
 * Update `candid` to v0.8.0.
--- a/icx-cert/Cargo.toml
+++ b/icx-cert/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "icx-cert"
-version = "0.20.0"
+version = "0.20.1"
 authors = ["DFINITY Stiftung <sdk@dfinity.org>"]
 edition = "2021"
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.22.0] - 2022-10-14
+
```
ericswanson-dfinity, line 12: Usually we change the release version separately from the changes that go into it.  Was there something that required changing the version in this PR?
```suggestion
```

icx-cert/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "icx-cert"
-version = "0.21.0"
+version = "0.22.0"
```
ericswanson-dfinity, line 3: Same for all Cargo.toml, with respect to changing the version number.
```suggestion
version = "0.21.0"
```

ericswanson-dfinity (changes requested):

CHANGELOG.md
mraszyk: It'd be great to have a release soon to enforce using the effective canister id for provisional_create_canister_with_cycles in our tests.

mraszyk: But we could also do this in a separate MR and only bump the minor version if you prefer (this PR is a breaking change though because read_state_raw takes only two arguments now).

ericswanson-dfinity: We also usually release a new agent right away after a change like this.  In this PR, please don't bump the major or minor versions.

icx-cert/Cargo.toml
mraszyk: Done.

CHANGELOG.md
mraszyk: I see. Done.

ericswanson-dfinity (approved):

PR #392, opened 2022-11-17
feat: Replace garcon with backoff

adamspofford-dfinity:
Implements SDK-824.

This additionally removes garcon from the public API; the backoff numbers dfx uses are good enough for anyone, and users who want a shorter timeout can use tokio's timeout function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Remove `garcon` from API.
+
 ## [0.22.0] - 2022-10-17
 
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Remove `garcon` from API.
+
```
ericswanson-dfinity, line 12: Since this is a breaking change to the API, can you say something here about what changes will be required?  In this case it should be simple since it's just `Waiter` parameters being removed/no longer necessary for callers.

ericswanson-dfinity (approved):

PR #414, opened 2023-03-22
feat: wasm-bindgen support

adamspofford-dfinity:
This PR makes ic-agent and ic-utils available in the browser through wasm-bindgen.


ericswanson-dfinity (commented):
What would it take to add any kind of test?

adamspofford-dfinity:
Selenium isn't necessary if you have a frontend, wasm-bindgen has its own unit test system. Writing a frontend that actually encases all the behavior, on the other hand, would be more work than writing the mock backend, and then you'd still need a backend.

PR #422, opened 2023-04-20
feat!: Handle rejected update calls after IC spec change.

### LINKED ISSUE #421: Handle spec change IC-1462
As part of the spec proposal [IC-1462](https://github.com/dfinity/interface-spec/pull/143), we need to do some changes in IC-Agent to handle HTTP responses with code 200 as that indicates there is an error message in the body. [The current implementation discards the body](https://github.com/dfinity/agent-rs/blob/42d96459852b6793bb9c44c297ed41f30b35b86a/ic-agent/src/agent/http_transport/reqwest_transport.rs#L265), but we need to map 200 status code to AgentError.

We should change the function "[execute](https://github.com/dfinity/agent-rs/blob/42d96459852b6793bb9c44c297ed41f30b35b86a/ic-agent/src/agent/http_transport/reqwest_transport.rs#L200)()" such that it handles responses with HTTP code 200 by returning an AgentError with the message extracted from the body response. So on line 238, we could for example have a check to if status == 200.

[IC-1462]: https://dfinity.atlassian.net/browse/IC-1462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

DSharifi:
# Description

This PR makes agent-rs handle reject responses for Calls after the IC Interface Specification change [IC-1462](https://github.com/dfinity/interface-spec/pull/143).

We need to do some changes in IC-Agent to handle HTTP responses with code 200 for `Call` requests as that indicates there is a CBOR encoded error message in the body.

It is a breaking change due to the enum variant `AgentErrror::ReplicaError` now being a tuple struct with attributes matching the new specification. 

Fixes #421

# How Has This Been Tested?
Two integration tests have been added in `agent_test.rs` to test the flow where a call request is rejected with status code 200.
Previous tests for queries have also been updated where `AgentError::ReplicaError` is expected as a response.

# Checklist:

[IC-1462]: https://dfinity.atlassian.net/browse/IC-1462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -1,6 +1,10 @@
 //! Errors that can occur when using the replica agent.
 
-use crate::{agent::status::Status, RequestIdError};
+use crate::{
+    agent::{replica_api::RejectResponse, status::Status},
+    RequestIdError,
+};
+// use crate::{agent::status::Status, RequestIdError};
 use ic_certification::Label;
 use leb128::read;
 use std::{
@@ -8,8 +12,6 @@ use std::{
     str::Utf8Error,
 };
 use thiserror::Error;
-
-use super::replica_api::RejectResponse;
 
 /// An error that occurred when using the agent.
 #[derive(Error, Debug)]
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -11,6 +11,8 @@ use thiserror::Error;
 
 use serde::Deserialize;
 
+use super::replica_api::{self, ReplicaError};
+
 /// An error that occurred when using the agent.
 #[derive(Error, Debug)]
 pub enum AgentError {
@@ -51,17 +53,8 @@ pub enum AgentError {
     PrincipalError(#[from] crate::export::PrincipalError),
 
     /// The replica rejected the message.
-    #[error(r#"The Replica returned an error: code {reject_code}, message: "{reject_message}""#)]
-    ReplicaError {
-        /// The [reject code](https://smartcontracts.org/docs/interface-spec/index.html#reject-codes) returned by the replica.
-        reject_code: u64,
-        /// The rejection message.
-        reject_message: String,
-    },
-
-    /// The replica rejected the message.
     #[error("Got a replica error: {0}")]
-    ReplicaErrorV2(ReplicaError),
+    ReplicaError(ReplicaError),
 
     /// The replica returned an HTTP error.
     #[error("The replica returned an HTTP Error: {0}")]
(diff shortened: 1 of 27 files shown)


ic-agent/src/agent/replica_api.rs
```diff
@@ -146,8 +146,50 @@ pub enum QueryResponse {
...
     Replied { reply: CallReply },
     #[serde(rename = "rejected")]
-    Rejected {
-        reject_code: u64,
-        reject_message: String,
-    },
+    Rejected(RejectResponse),
+}
+
+/// An HTTP error from the replica.
```
rumenov, line 152: This is not really an HTTP error. It is an IC execution error.

ic-agent/src/agent/agent_error.rs
```diff
@@ -193,6 +190,15 @@ impl PartialEq for AgentError {
     }
 }
 
+impl Display for RejectResponse {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        f.write_fmt(format_args!(
+            "Replica Error: reject code {:?}, reject message {}, error code {}",
+            self.reject_code, self.reject_message, self.reject_message,
```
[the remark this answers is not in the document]
DSharifi: Good spot! I'll fix it.

ic-agent/src/agent/http_transport/reqwest_transport.rs
```diff
@@ -236,7 +237,19 @@ impl ReqwestTransport {
             }
         }
 
-        if status.is_client_error() || status.is_server_error() {
+        // status == OK means we have an error message for call requests
+        // see https://internetcomputer.org/docs/current/references/ic-interface-spec#http-call
+        if status == StatusCode::OK && endpoint.ends_with("call") {
```
ericswanson-dfinity, line 242: It looks like a replica error (with reject_code, reject_message, and error_code) also come from an endpoint that ends in `/query`: https://internetcomputer.org/docs/current/references/ic-interface-spec#http-query

ericswanson-dfinity (changes requested):

ic-agent/src/agent/response_authentication.rs
```diff
@@ -82,24 +84,26 @@ pub(crate) fn lookup_rejection(
     let reject_code = lookup_reject_code(certificate, request_id)?;
     let reject_message = lookup_reject_message(certificate, request_id)?;
 
-    Ok(RequestStatusResponse::Rejected {
+    Ok(RequestStatusResponse::Rejected(RejectResponse {
         reject_code,
         reject_message,
-    })
+        error_code: None,
```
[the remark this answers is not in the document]
rumenov: No, this error is returned before any data is added to the replicated state.

ic-agent/src/agent/http_transport/reqwest_transport.rs
DSharifi: Yes, that is true. However, it's only for `/call` where an HTTP status code of 200 means that the response body is a replica error. The specification for `/query` has not changed in the spec change and the current approach for detecting replica errors is valid. There is an existing integration test covering the case where we have replica errors for query calls.

ic-agent/src/agent/agent_test.rs
```diff
@@ -85,10 +85,11 @@ async fn query_error() -> Result<(), AgentError> {
 #[cfg_attr(not(target_family = "wasm"), tokio::test)]
 #[cfg_attr(target_family = "wasm", wasm_bindgen_test)]
 async fn query_rejected() -> Result<(), AgentError> {
```
DSharifi, line 87: @ericswanson-dfinity Here is the integration test for mapping to replica errors on a rejected response.

ic-agent/src/agent/mod.rs
```diff
@@ -421,13 +422,9 @@ impl Agent {
...
             .and_then(|response| match response {
                 replica_api::QueryResponse::Replied { reply } => Ok(reply.arg),
-                replica_api::QueryResponse::Rejected {
-                    reject_code,
-                    reject_message,
-                } => Err(AgentError::ReplicaError {
-                    reject_code,
-                    reject_message,
-                }),
+                replica_api::QueryResponse::Rejected(response) => {
```
DSharifi, line 425: @ericswanson-dfinity Here you see how the logic of mapping rejected query calls to Replica Error stays the same.

ic-agent/src/agent/agent_test.rs
ericswanson-dfinity: Right, but I notice that this has `None` for the error_code.  Will error codes be available for queries?

icx/src/main.rs
```diff
@@ -567,14 +567,11 @@ async fn main() -> Result<()> {
...
-                        reject_message,
-                    } => {
+                    agent::RequestStatusResponse::Rejected(replica_eror) => {
                         bail!(
-                            r#"The Replica returned an error: code {}, message: "{}""#,
-                            reject_code,
-                            reject_message
+                            r#"The Replica returned an error: code {:?}, message: "{}""#,
+                            replica_eror.reject_code,
+                            replica_eror.reject_message
```
ericswanson-dfinity: Can we have this include the error code?

ic-agent/src/agent/agent_test.rs
DSharifi: Yes. I extended the test now to also include `Some("Error code")` in the error_code field.

icx/src/main.rs
DSharifi: Yes, that makes sense. I have included it now.

ericswanson-dfinity (approved):

ref-tests/tests/integration.rs
```diff
@@ -18,6 +18,8 @@ use ref_tests::{
     with_wallet_canister,
 };
```
rumenov, line 20: nit: you should merge with line 6.

ic-agent/src/agent/response_authentication.rs
```diff
@@ -3,6 +3,8 @@ use crate::{export::Principal, AgentError, RequestId};
 use ic_certification::{Certificate, Label, LookupResult};
 use std::str::from_utf8;
 
+use super::replica_api::{RejectCode, RejectResponse};
```
rumenov, line 6: nit: merge with line 1

ic-agent/src/agent/response.rs
```diff
@@ -1,3 +1,5 @@
+use super::replica_api::RejectResponse;
```
rumenov, line 1: nit: for consistency I think you should use 

"use crate::..."

ic-agent/src/agent/replica_api.rs
```diff
@@ -146,8 +147,53 @@ pub enum QueryResponse {
...
+    type Error = AgentError;
+
+    fn try_from(value: u64) -> Result<Self, AgentError> {
+        match value {
+            1 => Ok(RejectCode::SysFatal),
+            2 => Ok(RejectCode::SysTransient),
+            3 => Ok(RejectCode::DestinationInvalid),
+            4 => Ok(RejectCode::CanisterReject),
+            5 => Ok(RejectCode::CanisterError),
+            _ => Err(AgentError::MessageError(format!(
```
rumenov, line 193: I don't think the MessageError type is very appropriate here.

1.  The description reads -  #[error("A tool returned a string message error: {0}")]. There is actually no tool, but also it seems we never set this error in other place as well.
2. I think CallDataMismatch is more appropriate error.

If you agree I would suggest fixing this and at the same time even remove the AgentError::MessageError field. The error space is already too big for users of the IC agent to handle it gracefully. So no point in including error types that are not set.

ic-agent/src/agent/response.rs
DSharifi: Solved in #429

ref-tests/tests/integration.rs
DSharifi: Solved in #429

PR #424, opened 2023-04-21
fix: Fixed bug in conventional commits file

### LINKED ISSUE #423: PR Title checker fails for breaking canges.
The `Check PR Title` github job fails for PRs that are breaking changes following conventional commits by appending a `!` after the type/scope of the title.

DSharifi:
# Description

I copied the jobs from the [conventional commits configuration file from the SDK Repo.](https://sourcegraph.com/github.com/dfinity/sdk/-/blob/.github/workflows/conventional-commits.yml?L20-41)

This allows us to have PRs with a suffix of `!`, and should fix the job failing for the title of #422.
Fixes #423

# How Has This Been Tested?

N\A. The job is the same as used in https://sourcegraph.com/github.com/dfinity/sdk/-/blob/.github/workflows/conventional-commits.yml

# Checklist:

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -9,13 +9,22 @@ on:
 
 jobs:
   check:
+    name: conventional-pr-title:required
     runs-on: ubuntu-latest
     steps:
-      - uses: dfinity/conventional-pr-title-action@v2.2.3
-        with:
-          success-state: Title follows the specification.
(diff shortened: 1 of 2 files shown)


lwshang (approved):

smallstepman:
I believe changelog entry was unnecessary in this case - this is an internal change that relates only to CI and has no impact on the consumers of this code

PR #434, opened 2023-05-09
fix: multiple fixes to the agent-rs, most notably having 5 min limit on the retry loop while polling 

rumenov:
# Description

We want to use the public agent-rs as much as possible in the core protocol code and remove the existing custom solutions we have. For this purpose we may need to bridge some differences. For example there is no point in polling for a response more than 5 minutes since the replica will GC any ingress messages within 5 min anyway.

The MR also updates versions of some dependencies as well.

Fixes # (issue)

# How Has This Been Tested?

CI

# Checklist:

### DIFF
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -127,9 +127,9 @@ pub enum AgentError {
     #[error("The status response did not contain a root key.  Status: {0}")]
     NoRootKeyInStatus(Status),
 
-    /// Could not read the replica root key.
-    #[error("Could not read the root key")]
-    CouldNotReadRootKey(),
+    /// An operation failed due to a poisoned type.
+    #[error("Could not perform the requested operation due to a poisoned type")]
+    Poisoned(),
 
     /// The invocation to the wallet call forward method failed with an error.
     #[error("The invocation to the wallet call forward method failed with the error: {0}")]
--- a/ic-agent/src/agent/agent_error.rs
+++ b/ic-agent/src/agent/agent_error.rs
@@ -127,10 +127,6 @@ pub enum AgentError {
     #[error("The status response did not contain a root key.  Status: {0}")]
     NoRootKeyInStatus(Status),
 
-    /// An operation failed due to a poisoned type.
-    #[error("Could not perform the requested operation due to a poisoned type")]
-    Poisoned(),
-
     /// The invocation to the wallet call forward method failed with an error.
     #[error("The invocation to the wallet call forward method failed with the error: {0}")]
     WalletCallFailed(String),
--- a/ic-agent/src/agent/builder.rs
+++ b/ic-agent/src/agent/builder.rs
@@ -3,7 +3,6 @@ use crate::{
     AgentError, Identity, NonceFactory, NonceGenerator,
 };
 use std::sync::Arc;
-use std::time::Duration;
 
 /// A builder for an [`Agent`].
 #[derive(Default)]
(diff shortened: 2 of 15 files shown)


ic-agent/src/agent/agent_error.rs
```diff
@@ -127,9 +127,9 @@ pub enum AgentError {
     #[error("The status response did not contain a root key.  Status: {0}")]
     NoRootKeyInStatus(Status),
 
-    /// Could not read the replica root key.
-    #[error("Could not read the root key")]
-    CouldNotReadRootKey(),
+    /// An operation failed due to a poisoned type.
+    #[error("Could not perform the requested operation due to a poisoned type")]
+    Poisoned(),
```
adamspofford-dfinity, line 132: A poison state shouldn't be surfaced as an error at all. A call in the stack that does not have the ability to reset the mutex should receive it as a panic; it exists to propagate panics from one thread to another and if we do not provide the ability to recover from the other panic, the result should be unwrapped. Either way there should be no such error variant.

ic-agent/src/agent/mod.rs
```diff
@@ -263,19 +268,23 @@ impl Agent {
...
                 .unwrap_or_else(|| Duration::from_secs(300)),
-            root_key: Arc::new(RwLock::new(Some(IC_ROOT_KEY.to_vec()))),
+            root_key: Arc::new(RwLock::new(IC_ROOT_KEY.to_vec())),
             transport: config
                 .transport
                 .ok_or_else(AgentError::MissingReplicaTransport)?,
+            backoff_initial_interval: Duration::from_millis(500),
+            backoff_max_elapsed_time: Duration::from_secs(60 * 2),
+            backoff_max_interval: Duration::from_secs(1),
+            backoff_multiplier: 1.4,
```
adamspofford-dfinity, line 278: These attributes should also be surfaced in AgentBuilder.

ic-agent/src/agent/agent_error.rs
rumenov: Sure, this is even better. I just decided to keep the existing semantics but I do agree if we unwrap the lock it is better.

rumenov:
Thanks for the review @adamspofford-dfinity , can you please take another look

ic-agent/src/agent/builder.rs
```diff
@@ -83,12 +84,27 @@ impl AgentBuilder {
...
-                ..self.config
-            },
-        }
+    pub fn with_ingress_expiry(mut self, ingress_expiry: Option<std::time::Duration>) -> Self {
+        self.config.ingress_expiry = ingress_expiry;
+        self
+    }
+
+    /// Sets the initial backoff interval used for polling the IC when awaiting a response.
+    pub fn set_backoff_initial_interval(mut self, initial_interval: Duration) -> Self {
```
adamspofford-dfinity, line 93: These should be named `with_`, not `set_`.

ic-agent/src/agent/mod.rs
```diff
@@ -260,33 +265,20 @@ impl Agent {
...
-    /// NOTE: if you change the identity while having update calls in
-    /// flight, you will not be able to [Agent::poll] the status of these
-    /// messages.
-    pub fn set_identity<I>(&mut self, identity: I)
-    where
-        I: 'static + Identity,
-    {
-        self.identity = Arc::new(identity);
-    }
-
```
adamspofford-dfinity, line 289: unintentional deletion? set_identity is here on purpose.

rumenov: Actually it was intentional. Do we really need to change the identity and transport of an already existing agent ? Especially given the working message in the comments ?

adamspofford-dfinity: Yes. For the identity, a developer does not necessarily want to use the same identity for every call (and `poll` is little-used); for the transport, it is used in `dfx canister sign` specifically to wrap an existing agent.

rumenov: ok, I reverted it for now. I can imagine that a user would like to set a new identity for different requests. However, setting a different transport may be a bit off. I would imagine if someone needs a new transport they would just create a new agent.

adamspofford-dfinity (approved):

PR #435, opened 2023-05-10
fix: put wasm-bindgen build target behind a feature.

DSharifi:
# Description
It is necessary to put the `wasm-bindgen` build target behind a feature in order to use it in the [ic repo](https://github.com/dfinity/ic). Due to bazel and cargo's feature unification, some transitive dependencies such as `getrandom`, `backoff`, and `time` crates were unified which were breaking our builds in the main IC-repo, making it impossible to use the current state of `ic-agent`.

As a consequence, we also have to update the GitHub workflows to include `--features wasm-bindgen`.

Fixes # (issue)

# How Has This Been Tested?
`cargo test`.

# Checklist:

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -56,7 +56,7 @@ const IC_ROOT_KEY: &[u8; 133] = b"\x30\x81\x82\x30\x1d\x06\x0d\x2b\x06\x01\x04\x
 #[cfg(not(target_family = "wasm"))]
 type AgentFuture<'a, V> = Pin<Box<dyn Future<Output = Result<V, AgentError>> + Send + 'a>>;
 
-#[cfg(target_family = "wasm")]
+#[cfg(all(target_family = "wasm", feature = "wasm-bindgen"))]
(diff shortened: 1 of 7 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -56,7 +56,7 @@ const IC_ROOT_KEY: &[u8; 133] = b"\x30\x81\x82\x30\x1d\x06\x0d\x2b\x06\x01\x04\x
 #[cfg(not(target_family = "wasm"))]
 type AgentFuture<'a, V> = Pin<Box<dyn Future<Output = Result<V, AgentError>> + Send + 'a>>;
 
-#[cfg(target_family = "wasm")]
+#[cfg(all(target_family = "wasm", feature = "wasm-bindgen"))]
```
adamspofford-dfinity, line 59: This one should be reverted - it deals with interface rather than implementation. The effect of applying the wasm-bindgen feature should be invisible in the docs.
```suggestion
#[cfg(target_family = "wasm")]
```

adamspofford-dfinity (approved):

PR #453, opened 2023-08-03
feat: Unknown lookup status of a path in certificate yields AgentError

mraszyk:
# Description

Absence proofs have been recently enabled on the IC (see [MR](https://github.com/dfinity/ic/commit/2ac0fe91cd737977bf039c55502a31b363884c69)) and thus the IC is going to produce certificate whose lookup returns `Absent` for paths whose values are missing in the state tree. This PR reflects this change in the agent.

### DIFF
--- a/ic-agent/src/agent/response_authentication.rs
+++ b/ic-agent/src/agent/response_authentication.rs
@@ -67,7 +67,7 @@ pub(crate) fn lookup_request_status<Storage: AsRef<[u8]>>(
     ];
     match certificate.tree.lookup_path(&path_status) {
         LookupResult::Absent => Ok(RequestStatusResponse::Unknown),
-        LookupResult::Unknown => Ok(RequestStatusResponse::Unknown),
+        LookupResult::Unknown => Err(LookupPathUnknown(path_status)),
         LookupResult::Found(status) => match from_utf8(status)? {
             "done" => Ok(RequestStatusResponse::Done),
             "processing" => Ok(RequestStatusResponse::Processing),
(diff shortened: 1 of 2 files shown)


ic-agent/src/agent/response_authentication.rs
```diff
@@ -67,7 +67,7 @@ pub(crate) fn lookup_request_status<Storage: AsRef<[u8]>>(
     ];
     match certificate.tree.lookup_path(&path_status) {
         LookupResult::Absent => Ok(RequestStatusResponse::Unknown),
-        LookupResult::Unknown => Ok(RequestStatusResponse::Unknown),
+        LookupResult::Unknown => Err(LookupPathUnknown(path_status.to_vec())),
```
ericswanson-dfinity, line 70: To see if I understand:
- Before https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/13886, IC would return LookupResult::Unknown
- After, IC will return LookupResult::Absent

Note that Agent::wait() will exit immediately if lookup_request_status returns Err(LookupPathUnknown).  Does this mean we need to wait to publish an agent version with this change until the IC is running a replica that incorporates MR13886?

ericswanson-dfinity (changes requested):

ic-agent/src/agent/response_authentication.rs
mraszyk: Indeed, the above two bullet points are correct.

> we need to wait to publish an agent version with this change until the IC is running a replica that incorporates MR13886

yes, indeed

ericswanson-dfinity (changes requested):
Looks good.  Let's hold off on merging this until the relevant replica update has been executed.  Otherwise, we wouldn't be able to release agent-rs in the meantime with other changes.

mraszyk:
The replica [release](https://dashboard.internetcomputer.org/release/3bcccef07408921fe849c92dd2437adc157ef9c3) containing absence proofs has been rolled out on all subnets. So we can proceed with this MR.

ericswanson-dfinity (approved):

PR #463, opened 2023-08-24
feat: sign delegations and arbitrary data

adamspofford-dfinity:
The two things an identity is expected to sign in the IC spec are content maps and delegations; this PR adds delegations where ic-agent currently only supports the content map.

This PR also partially reverts the constriction imposed in #443 to re-allow arbitrary garbage to be signed without being a valid content map; it, like delegations, is an explicitly *optional* feature of the Identity trait which not all implementations support.

Finally a `public_key` function is added so `sign_arbitrary(&[])` is not required in order to get the unhashed public key for the identity.

### DIFF
--- a/ic-agent/src/identity/mod.rs
+++ b/ic-agent/src/identity/mod.rs
@@ -49,6 +49,8 @@ pub trait Identity: Send + Sync {
 
     /// Sign a delegation to let another key be used to authenticate [`sender`](Identity::sender).
     ///
+    /// Not all `Identity` implementations support this operation.
+    ///
     /// Implementors should call `content.signable()` for the actual bytes that need to be signed.
     fn sign_delegation(&self, content: &Delegation) -> Result<Signature, String> {
         let _ = content; // silence unused warning
@@ -83,13 +85,15 @@ pub struct Delegation {
     pub senders: Option<Vec<Principal>>,
 }
 
+const IC_REQUEST_DELEGATION_DOMAIN_SEPARATOR: &[u8] = b"\x1Aic-request-auth-delegation";
+
 impl Delegation {
     /// Returns the signable form of the delegation, by running it through [`to_request_id`]
     /// and prepending `\x1Aic-request-auth-delegation` to the result.
     pub fn signable(&self) -> Vec<u8> {
         let hash = to_request_id(self).unwrap();
         let mut bytes = Vec::with_capacity(59);
-        bytes.extend_from_slice(b"\x1Aic-request-auth-delegation");
+        bytes.extend_from_slice(IC_REQUEST_DELEGATION_DOMAIN_SEPARATOR);
         bytes.extend_from_slice(hash.as_slice());
         bytes
     }
--- a/ic-agent/src/identity/mod.rs
+++ b/ic-agent/src/identity/mod.rs
@@ -1,5 +1,7 @@
 //! Types and traits dealing with identity across the Internet Computer.
-use crate::{agent::EnvelopeContent, export::Principal};
+use crate::{agent::EnvelopeContent, export::Principal, to_request_id};
+
+use serde::{Deserialize, Serialize};
 
 pub(crate) mod anonymous;
 pub(crate) mod basic;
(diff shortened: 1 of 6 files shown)


ericswanson-dfinity (changes requested):
Please describe this change in the changelog

ic-agent/src/identity/mod.rs
```diff
@@ -24,19 +26,75 @@ pub struct Signature {
...
+    ///
+    /// Implementors should call `content.signable()` for the actual bytes that need to be signed.
+    fn sign_delegation(&self, content: &Delegation) -> Result<Signature, String> {
+        let _ = content; // silence unused warning
+        Err(String::from("unsupported"))
+    }
+
+    /// Sign arbitrary bytes.
+    ///
+    /// Not all `Identity` implementations support this operation.
```
chenyan-dfinity, line 62: Please document that all identities, except the anonymous identity, implemented in `ic-agent` supports this method.

ic-agent/src/identity/mod.rs
```diff
@@ -24,19 +26,75 @@ pub struct Signature {
...
 /// identities used.
 pub trait Identity: Send + Sync {
     /// Returns a sender, ie. the Principal ID that is used to sign a request.
     ///
     /// Only one sender can be used per request.
     fn sender(&self) -> Result<Principal, String>;
 
+    /// Produce the public key commonly returned in [`Signature`].
+    ///
+    /// May return `None` when [`sign`](Identity::sign) would return `Some`.
```
chenyan-dfinity, line 42: ditto

ic-agent/src/identity/mod.rs
```diff
@@ -24,19 +26,75 @@ pub struct Signature {
...
 pub trait Identity: Send + Sync {
     /// Returns a sender, ie. the Principal ID that is used to sign a request.
     ///
     /// Only one sender can be used per request.
     fn sender(&self) -> Result<Principal, String>;
 
+    /// Produce the public key commonly returned in [`Signature`].
+    ///
+    /// May return `None` when [`sign`](Identity::sign) would return `Some`, in non-`ic-agent` identities.
+    fn public_key(&self) -> Option<Vec<u8>>;
```
chenyan-dfinity, line 43: Do we want to provide a default implementation to return `None`?

adamspofford-dfinity: No. The reason that is there is for signatures that differ in their public keys depending on the content being signed, *not* because it should be considered an optional function.

chenyan-dfinity: Let's document that. It's unlike when this function returns None.

adamspofford-dfinity: Actually I think a better option is to remove it since the use-case I was envisioning is better designed differently.

chenyan-dfinity (approved):
LGTM. I tried to patch this to the ic repo and everything is working!

ic-agent/src/identity/mod.rs
```diff
@@ -24,19 +26,75 @@ pub struct Signature {
...
+    ///
+    /// Implementors should call `content.signable()` for the actual bytes that need to be signed.
+    fn sign_delegation(&self, content: &Delegation) -> Result<Signature, String> {
+        let _ = content; // silence unused warning
+        Err(String::from("unsupported"))
+    }
+
+    /// Sign arbitrary bytes.
+    ///
+    /// Not all `Identity` implementations support this operation, though all `ic-agent` implementations do.
```
chenyan-dfinity, line 62: Can you explain why some implementation won't have sign_arbitrary? Eventually, we will get the raw bytes from the envelope, right?

chenyan-dfinity: Please document that. It's not easy for the user to know when to implement `sign_arbitrary` since it's optional.

adamspofford-dfinity: I do not think 'implement all the operations you can' needs explicit documentation.

chenyan-dfinity: As a consumer of the API, I need some guideline for when I can expect the optional methods to work without checking the individual implementation. Your reply makes sense, and putting it in the doc will help other users too.

adamspofford-dfinity: There is no such general guideline; if an implementation has a reason it cannot support that operation, the reason is specific to the implementation.

ericswanson-dfinity (approved):

PR #465, opened 2023-08-28
fix: remove argument builder from ic-utils

chenyan-dfinity:
# Description

* Remove argument builder from `ic-utils` to prevent requiring `Send` for `candid::Type`. It's never correct to use `IDLBuilder` in the argument builder, as Candid uses `thread_local` to resolve `TypeId`, which is not `Send`.
* Reverts #454 

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/ic-utils/src/canister.rs
+++ b/ic-utils/src/canister.rs
@@ -131,29 +131,38 @@ impl<'agent> Canister<'agent> {
 
 /// A buffer to hold canister argument blob.
 #[derive(Debug)]
-pub struct Argument(Vec<u8>);
+pub struct Argument(Result<Vec<u8>, AgentError>);
 
 impl Argument {
-    /// Set an IDL Argument, replacing existing arguments.
+    /// Set an IDL Argument, replacing existing arguments. If the current value is an error, will do nothing.
     pub fn push_idl_arg<A: CandidType>(&mut self, arg: A) {
-        self.0 = Encode!(&arg).unwrap();
+        if self.0.is_ok() {
+            self.0 = Encode!(&arg).map_err(|e| e.into());
+        }
     }
 
-    /// Set an IDLValue Argument, replacing existing arguments.
+    /// Set an IDLValue Argument, replacing existing arguments. If the current value is an error, will do nothing.
     pub fn push_value_arg(&mut self, arg: IDLValue) {
-        let mut builder = IDLBuilder::new();
-        builder.value_arg(&arg).unwrap();
-        self.0 = builder.serialize_to_vec().unwrap();
+        if self.0.is_ok() {
+            let mut builder = IDLBuilder::new();
+            let result = builder
+                .value_arg(&arg)
+                .and_then(|builder| builder.serialize_to_vec())
+                .map_err(|e| e.into());
+            self.0 = result;
+        }
     }
 
     /// Set the argument as raw, replacing existing arguments.
     pub fn set_raw_arg(&mut self, arg: Vec<u8>) {
-        self.0 = arg;
+        if self.0.is_ok() {
+            self.0 = Ok(arg);
+        }
     }
 
     /// Return the argument blob.
     pub fn serialize(self) -> Result<Vec<u8>, AgentError> {
-        Ok(self.0)
+        self.0
     }
 
     /// Resets the argument to an empty message.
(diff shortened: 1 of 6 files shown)


adamspofford-dfinity (changes requested):
This entirely changes the semantics of the `with_arg` functions, making it impossible to pass multiple arguments and silently breaking any code that already does.

chenyan-dfinity:
@eric-swanson What do you think?

ericswanson-dfinity:
> @eric-swanson What do you think?

How about this?

Make `.with_arg()` report an error (either panic in `with_arg()`, or have `build()` return an error) if called more than once for a builder.  The error can tell what you should do ("`use CallBuilder::set_raw_arg(candid::Encode!(arg1, arg2)?)`") instead.

chenyan-dfinity:
Sounds good! Will do.

ericswanson-dfinity:
> Sounds good! Will do.

Are you satisfied with that @adamspofford-dfinity ?  The only other thing I can think of would be for `.with_arg` to return a Builder that doesn't have any of the argument-adding methods, in order to catch it at compile time.

adamspofford-dfinity:
I would like the alternative to be an ArgumentEncoder function the way `with_arg` is a CandidType function; a name that springs to mind is `with_args`. Otherwise, yes, that works.

chenyan-dfinity:
Sounds good.

lwshang:
> Reverts https://github.com/dfinity/agent-rs/pull/454
Will we be able to remove the `arg_type` feature of candid crate?

chenyan-dfinity:
> Will we be able to remove the arg_type feature of candid crate?

Yep

ic-utils/src/interfaces/wallet.rs
```diff
@@ -61,7 +62,7 @@ where
     where
         Argument: CandidType + Sync + Send,
     {
-        self.arg.push_idl_arg(arg);
+        self.arg.set_idl_arg(arg);
         self
     }
```
adamspofford-dfinity, line 68: This should also have a with_args overload.

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -395,7 +392,7 @@ impl<'agent, 'canister: 'agent> InstallCodeBuilder<'agent, 'canister> {
         mut self,
         arg: Argument,
     ) -> InstallCodeBuilder<'agent, 'canister> {
-        self.arg.push_idl_arg(arg);
+        self.arg.set_idl_arg(arg);
         self
     }
```
adamspofford-dfinity, line 398: This should also have a with_args overload.

chenyan-dfinity:
PTAL

ic-utils/src/canister.rs
```diff
@@ -272,28 +223,32 @@ impl<'agent, 'canister: 'agent> SyncCallBuilder<'agent, 'canister> {
...
         Argument: CandidType + Sync + Send,
     {
-        self.arg.push_idl_arg(arg);
+        self.arg.set_idl_arg(arg);
+        self
+    }
+    /// Set the argument with multiple arguments as tuple. Can be called at most once.
+    pub fn with_args(mut self, tuple: impl ArgumentEncoder) -> SyncCallBuilder<'agent, 'canister> {
+        if self.arg.0.is_some() {
+            panic!("argument is being set for more than once");
```
ericswanson-dfinity, line 237:
```suggestion
            panic!("argument is being set more than once");
```

ic-utils/src/canister.rs
```diff
@@ -353,18 +308,24 @@ impl<'agent, 'canister: 'agent> AsyncCallBuilder<'agent, 'canister> {
...
         Argument: CandidType + Sync + Send,
     {
-        self.arg.push_idl_arg(arg);
+        self.arg.set_idl_arg(arg);
+        self
+    }
+    /// Set the argument with multiple arguments as tuple. Can be called at most once.
+    pub fn with_args(mut self, tuple: impl ArgumentEncoder) -> AsyncCallBuilder<'agent, 'canister> {
+        if self.arg.0.is_some() {
+            panic!("argument is being set for more than once");
```
ericswanson-dfinity, line 322:
```suggestion
            panic!("argument is being set more than once");
```

adamspofford-dfinity (approved):

PR #478, opened 2023-09-30
feat: Rewrite request ID calculator

adamspofford-dfinity:
This PR allows request IDs to include nested maps, which necessitated a rewrite of the underlying algorithm. Required for replica-signed queries.

### DIFF
--- a/ic-transport-types/src/request_id.rs
+++ b/ic-transport-types/src/request_id.rs
@@ -3,7 +3,7 @@
 //!
 //! A request ID is a SHA256 hash of the request's body. See
 //! [Representation-independent Hashing of Structured Data](https://internetcomputer.org/docs/current/references/ic-interface-spec#hash-of-map)
-/// from the IC spec for the method of calculation.
+//! from the IC spec for the method of calculation.
 use error::RequestIdFromStringError;
 use serde::{
     ser::{
@@ -60,7 +60,7 @@ impl RequestId {
     }
 
     /// Returns the signable form of the request ID, by prepending `"\x0Aic-request"` to it,
-    /// for use in [`Identity::sign`](crate::identity::Identity::sign).
+    /// for use in [`Identity::sign`](https://docs.rs/ic-agent/latest/ic_agent/trait.Identity.html#tymethod.sign).
     pub fn signable(&self) -> Vec<u8> {
         let mut signable = Vec::with_capacity(43);
         signable.extend_from_slice(IC_REQUEST_DOMAIN_SEPARATOR);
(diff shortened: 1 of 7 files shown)


krpeacock:
This PR removes a lot of comments - are they no longer applicable with the new changes?

ic-transport-types/src/request_id.rs
```diff
@@ -854,20 +652,154 @@ mod tests {
...
+        let outer = Outer {
+            foo: Inner {
+                baz: "hello",
+                quux: 3,
+            },
+            bar: b"world",
+        };
+        assert_eq!(
+            hex::encode(to_request_id(&outer).unwrap().0),
+            "3d447339cc0c2b894ee215c8141770bf4b86c72b6c37d9873213a786ec7f9f31"
```
ericswanson-dfinity, line 677: Could we add tests like this (compare computed hash with specific expected hash) for Tuple, Newtype, and any other previously unsupported types?

lwshang:
The Request ID example in interface-spec has changed. (Two extra fields)
IMO, we'd better update the corresponding tests: `public_spec_example` and `public_spec_example_api_client`.

```rust
    #[test]
    fn public_spec_example() {
        #[derive(Serialize)]
        struct PublicSpecExampleStruct {
            request_type: &'static str,
            sender: Principal,
            ingress_expiry: u64,
            canister_id: Principal,
            method_name: &'static str,
            #[serde(with = "serde_bytes")]
            arg: Vec<u8>,
        }
        let data = PublicSpecExampleStruct {
            request_type: "call",
            sender: Principal::anonymous(),
            ingress_expiry: 1685570400000000000u64,
            canister_id: Principal::try_from(&vec![0, 0, 0, 0, 0, 0, 0x04, 0xD2]).unwrap(), // 1234 in u64
            method_name: "hello",
            arg: b"DIDL\x00\xFD*".to_vec(),
        };

        // Hash taken from the example on the public spec.
        let request_id = to_request_id(&data).unwrap();
        assert_eq!(
            hex::encode(request_id.0),
            "1d1091364d6bb8a6c16b203ee75467d59ead468f523eb058880ae8ec80e2b101"
        );
    }
```

The implementation looks great. And the comments are sufficient.

As Eric suggested, more tests will be great for coverage. The existing test vectors were generated with `ic-ref`. So it might be more efficient to get some help from `ic-ref` folks.

adamspofford-dfinity:
@krpeacock Either no longer applicable, capable of being made more concise by locating closer to the code they actually document, or were originally copied verbatim from Serializer's docs and thus pointless.

lwshang (approved):

PR #479, opened 2023-10-02
chore: Raise MSRV to 1.70.0

adamspofford-dfinity:
This PR raises the MSRV to 1.70 following several dependencies doing the same. Particular to 1.70 is the support for the sparse Cargo registry, to accelerate this update in particular. It also removes the distinction between the CI version and the local dev version, letting CI pick the version up from rust-toolchain.toml.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -23,10 +23,6 @@ jobs:
             ~/.cargo/git
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-1
-      - name: Install Rust
-        uses: ructions/toolchain@v2
-        with:
-          profile: minimal
 
       - name: Provision Darwin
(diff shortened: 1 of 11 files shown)


.github/workflows/e2e.yml
```diff
@@ -25,10 +24,9 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-1
       - name: Install Rust
-        run: |
-          rustup update ${{ matrix.rust }} --no-self-update
-          rustup default ${{ matrix.rust }}
-          rustup component add rustfmt
+        uses: ructions/toolchain@v2
```
lwshang, line 27: Explicitly install Rust toolchain is not necessary anymore. GitHub actions runner has `rustup` by default. It will use the version/components in `rust-toolchain.toml` whenever call `cargo` commands.

ericswanson-dfinity (approved):

PR #480, opened 2023-10-04
feat: replica-signed queries

adamspofford-dfinity:
This PR implements support for the node signatures provided in query calls by looking up the subnet node keys in the state tree. This provides resilience against rogue boundary nodes; clients will know whether the response came from the canister's subnet.

Implements SDK-1213 and SDK-1215.


ic-agent/src/agent/builder.rs
```diff
@@ -87,4 +87,11 @@ impl AgentBuilder {
         self.config.ingress_expiry = ingress_expiry;
         self
     }
+
+    /// Allows disabling query signature verification. Query signatures improve resilience but require
+    /// a separate read-state call to fetch node keys.
```
ericswanson-dfinity, line 92: Should we mention that once a canister -> subnet mapping is looked up, it is cached and never invalidated?  If canisters can change subnets, then this is something that someone who enables verify_query_signatures should be aware of.

ic-agent/src/agent/builder.rs
```diff
@@ -87,4 +87,11 @@ impl AgentBuilder {
         self.config.ingress_expiry = ingress_expiry;
         self
     }
+
+    /// Allows disabling query signature verification. Query signatures improve resilience but require
```
ericswanson-dfinity, line 91: It looks like the default (`verify_query_signatures: false,`) is to disable query signature verification.  But this comment implies the opposite ("Allows disabling..").  


```suggestion
    /// Allows configuration of query signature verification, which is disabled by default. Query signatures improve resilience but require
```

adamspofford-dfinity: good catch - it's actually supposed to be enabled by default.

ic-agent/src/agent/builder.rs
adamspofford-dfinity: I had thought it was transitively invalidated by the subnet information being invalidated, since it's just an index, but hadn't considered the possibility of the subnet's ranges changing.

ericswanson-dfinity: I think [subnet splitting](https://github.com/dfinity/portal/pull/1966) would be an event that would change the subnet of some canisters

ericswanson-dfinity (approved):

adamspofford-dfinity:
Not to be merged, pending review by @roelstorms / @robin-kunzler.

ic-agent/Cargo.toml
```diff
@@ -16,7 +16,9 @@ include = ["src", "Cargo.toml", "../LICENSE", "README.md"]
 
 [dependencies]
 backoff = "0.4.0"
+cached = { version = "0.46", features = ["ahash"], default-features = false }
 candid = { workspace = true }
+ed25519-dalek = { version = "2" }
```
roelstorms, line 21: Is there a reason not to use the [ed25519 library used in the ICs crypto component](https://github.com/dfinity/ic/blob/master/rs/crypto/internal/crypto_lib/basic_sig/ed25519/src/api.rs#L162)?

ic-agent/src/agent/agent_config.rs
```diff
@@ -23,6 +25,7 @@ impl Default for AgentConfig {
             identity: Arc::new(AnonymousIdentity {}),
             ingress_expiry: None,
             transport: None,
+            verify_query_signatures: true,
```
roelstorms, line 28: For verify_query_signatures we set the default to true in the config.
For ingress_expiry, we set the default to 300 seconds in Agent::new.

Just wondering why it's done differently.

ic-agent/src/agent/response_authentication.rs
```diff
@@ -132,25 +137,69 @@ pub(crate) fn lookup_reply<Storage: AsRef<[u8]>>(
...
+        if path[1].as_bytes() != b"public_key" {
+            continue;
+        }
+        if path.len() > 2 {
+            return Err(AgentError::LookupPathError(
+                path.into_iter()
+                    .map(|label| label.as_bytes().to_vec().into())
+                    .collect(),
+            ));
+        }
```
roelstorms, line 174: We are being rather strict as to what all paths under /node look like.
If the spec changes in the future and a node can respond with other paths, this code would fail.

Should it return an error or simply skip paths which do not adhere to the /<node_id>/public_key format we expect?

ic-agent/Cargo.toml
adamspofford-dfinity: None of the internal crypto crates are published to crates.io.

ic-transport-types/src/lib.rs
```diff
@@ -122,9 +122,73 @@ pub enum QueryResponse {
...
+                error_code: reject.error_code.as_ref(),
+                reject_code: reject.reject_code,
+                reject_message: &reject.reject_message,
+                request_id,
+                timestamp,
+            },
+        };
+        let mut signable = Vec::with_capacity(44);
+        signable.extend_from_slice(b"\x0Bic-response");
+        signable.extend_from_slice(to_request_id(&response).unwrap().as_slice());
```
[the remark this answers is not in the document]
adamspofford-dfinity: It has always been exported under this name, even when used for other things. 🤷

ic-agent/src/agent/response_authentication.rs
adamspofford-dfinity: It would fail if `public_key` ever became a branch node instead of a leaf node, but there is already nothing to do about that since then it could not contain a value. And if it receives any path that is not `public_key`, it skips over it instead of failing.

ic-agent/src/agent/agent_config.rs
adamspofford-dfinity: We don't guarantee that ingress_expiry's default has to be representable as a single value. For example, we could restrict minute rounding behavior to only apply when ingress_expiry is not explicitly defined. But verifying query signatures is a binary state - either it happens or it doesn't.

ic-agent/src/agent/mod.rs
```diff
@@ -416,14 +426,73 @@ impl Agent {
...
+                let pubkey = VerifyingKey::from_bytes(node_key[12..].try_into().unwrap()).unwrap();
+                let sig = Signature::from_bytes(
+                    signature.signature[..]
+                        .try_into()
+                        .map_err(|_| AgentError::CertificateVerificationFailed())?,
+                );
+
+                if pubkey.verify(&signable, &sig).is_err() {
+                    return Err(AgentError::CertificateVerificationFailed());
+                }
```
ShuoWangNSL, line 484: Hi @andreacerulli  Shall we have members from the crypto team to have a look at the ed25519_dalek library used here as part of the [ticket]( https://dfinity.atlassian.net/browse/CRP-2203) ?

roelstorms: It looks like the crypto team is using this library themselves: https://docs.rs/ed25519-consensus/latest/ed25519_consensus/

Maybe worth getting their advice. According to Robin Kunzler they had a good reason to choose this lib.

ic-agent/src/agent/response_authentication.rs
roelstorms: If there is ever another 1 length path under /node this would return AgentError::CertificateNotAuthorized(), right?

Anyway, not a concrete security risk at this time. Might need to be updated if the spec ever changes which might be overlooked at that point.

adamspofford-dfinity: Given that it goes `node/<node id>/public_key`, I do not think there is any risk of anything else going under `node`, since the whole point is to treat it as a hash-map and you could not otherwise detect whether you had a node ID or not.

ic-agent/src/agent/mod.rs
```diff
@@ -782,6 +863,31 @@ impl Agent {
...
+            .subnet_key_cache
+            .lock()
+            .unwrap()
+            .get_subnet_by_canister(canister);
+        if let Some(subnet) = subnet {
+            Ok(subnet)
+        } else {
+            let cert = self
+                .read_state_raw(vec![vec!["subnet".into()]], *canister)
+                .await?;
```
ShuoWangNSL, line 881: Usually, the canister ranges check is performed within check_delegation.
However, if the certificate is from the root subnet, the delegation is absent. In the context of replica-signed queries, we need to perform extra check of the canister ranges for the root subnet, .
```
if cert.delegation.is_none(){
    let canister_range = lookup_value(&cert.tree, canister_range_lookup)?;
    let ranges: Vec<(Principal, Principal)> =
        serde_cbor::from_slice(canister_range).map_err(AgentError::InvalidCborData)?;
    if !principal_is_within_ranges(&effective_canister_id, &ranges[..]) {
        // the certificate is not authorized to answer calls for this canister
        return Err(AgentError::CertificateNotAuthorized());
    }
}
                
```

ShuoWangNSL: The design slides and @roelstorms 's [comments](https://docs.google.com/presentation/d/1xbobmOhplo-OwjP41JY5MhpI7c-peyJPtNE-v_-DJKo/edit?disco=AAAA-gP82Ik) may be referenced for more context.

ic-agent/src/agent/agent_config.rs
adamspofford-dfinity: #481

ic-agent/src/agent/mod.rs
```diff
@@ -782,6 +863,31 @@ impl Agent {
...
+        &self,
+        canister: &Principal,
+    ) -> Result<Arc<Subnet>, AgentError> {
+        let subnet = self
+            .subnet_key_cache
+            .lock()
+            .unwrap()
+            .get_subnet_by_canister(canister);
+        if let Some(subnet) = subnet {
+            Ok(subnet)
```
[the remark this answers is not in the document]
ShuoWangNSL: It seems it is automatically handled by TimedCache.  Let me take a look at how it behaves.

ic-agent/src/agent/mod.rs
```diff
@@ -1020,6 +1126,72 @@ pub fn signed_request_status_inspect(
...
+            .get(canister)
+            .and_then(|subnet_id| self.subnets.cache_get(subnet_id).cloned())
+            .filter(|subnet| subnet.canister_ranges.contains(canister))
+    }
+
+    fn insert_subnet(&mut self, subnet_id: Principal, subnet: Arc<Subnet>) {
+        self.subnets.cache_set(subnet_id, subnet.clone());
+        for range in subnet.canister_ranges.iter() {
+            self.canister_index.insert(range.clone(), subnet_id);
+        }
```
ShuoWangNSL, line 1154: In the case of canister migration(where a canister is moved from subnet A to B) and subnet splitting(where subnet A is split into subnet A’ and B), the subnet’s canister ranges can change.

Suppose the agent talks to canister_1 and canister_2 in the subnet A. Then canister_2 is moved to B. If the agent keeping making queries to canister_1 then the cache for subnet A will always be there when the agent makes queries to canister_2. Thus a new remote request for fetching node keys for canister_2 will not be triggered and in turn fail all queries to canister_2.

I am not sure how RangeInclusiveMap exactly works. If stale ranges for A can automatically be removed and then new ones for A are inserted when we refetch node keys for A, this will solve the problem that canister_2 is always mapped to subnet A. If the agent does not make queries will canister_1 anymore, it is true that the canister_index is stale. However, it won't be a problem for canister_2, because the cache miss will trigger remote fetch of subnet as it will work as long as overlapped ranges between A and B can be overridden.

Do you think this solution is feasible? In terms of code, what do you think of first removing all existing ranges mapped to the subnet_id and then do the for loop to insert new ones?

cc @ericswanson-dfinity @adamspofford-dfinity @roelstorms

ShuoWangNSL: @krpeacock I see you use a direct mapping from canister to subnet in agent-js for the cache. 
In the case of canister ranges changes,  the mapping may be invalid. Updating the direct mapping from the canister to the new subnet should be easy, right?

ic-transport-types/src/lib.rs
roelstorms, line 181: to_request_id could panic, should an agent error be returned instead? 
Now, a malicious replica can crash the software using agent-rs.

One way this might panic is if the RequestIdSerializer is handed a QueryResponseSignable::Rejected with an error code outside the range 1-5.

ic-agent/src/agent/mod.rs
roelstorms: I'm wondering if we only need to do this in case of RSQ:

> In the context of replica-signed queries, we need to perform extra check of the canister ranges for the root subnet,

I'd argue that this might be a general issue that a node on the root subnet could respond with the root cert (and thus root state tree) while requesting the state for a canister on another subnet.

In our case the issue is that we have the list of authorative nodes in that state and the root subnet nodes shouldn't be authoritive for RSQ on other subnets. But maybe there are other scenario's where this could be an issue. 

Is it worth mitigating this in verify(). In case there is no delegation, the effective canister id needs to be within the range of root subnet. I'm not sure if the root subnet also publishes it's own canister ranges in it's own certificate?

ShuoWangNSL: > I'm not sure if the root subnet also publishes it's own canister ranges in it's own certificate

It is a special case for replica-signed queries where agents request the `/subnet` path and thus it contains `/subnet/<subnet-id>/canister_ranges` in the cert directly. 

In other cases of read_state call for other paths, `/subnet/<subnet-id>/canister_ranges` may not be present in the cert and thus moving the logic to `verify()` may not work.

ic-agent/src/agent/mod.rs
```diff
@@ -416,14 +426,73 @@ impl Agent {
...
+                }
+                let pubkey = VerifyingKey::from_bytes(node_key[12..].try_into().unwrap()).unwrap();
+                let sig = Signature::from_bytes(
+                    signature.signature[..]
+                        .try_into()
+                        .map_err(|_| AgentError::CertificateVerificationFailed())?,
+                );
+
+                if pubkey.verify(&signable, &sig).is_err() {
+                    return Err(AgentError::CertificateVerificationFailed());
```
ShuoWangNSL, line 483: Shall we introduce another AgentError type `QueryResponseSignatureVerificationFailed`? This can give users more information as it is a new case introduced by the replica-signed queries feature.  `CertificateVerificationFailed` may be misleading and make users think the failure comes from a read_state call.

ic-agent/Cargo.toml
roelstorms: The crypto team uses this library underneath: https://crates.io/crates/ed25519-consensus
According to Robin Kunzler, they had a very good reason to do so. Could you check with them if the library you intend to use is a good replacement?

ic-transport-types/src/lib.rs
adamspofford-dfinity: Where are you seeing that this would cause a panic?

ic-agent/src/agent/mod.rs
adamspofford-dfinity: @ShuoWangNSL When a shrunken range for A is reinserted, it will indeed not update the rangemap index to account for canister_2 not being in it. However, there's a final check that canister_2 is in A's range, and if this check fails, it is still a cache miss rather than an error, which will trigger looking up B and updating the index.

adamspofford-dfinity: It could be written to remove old indexes, but since there's going to be range-related cache misses *anyway*, I'd rather they result in more hash lookups than more network requests.

ShuoWangNSL: > there's a final check that canister_2 is in A's declared range, and when this check fails, it is still a cache miss rather than an error, which will trigger looking up B and updating the index.

I got it. If the cache of subnet A expires, not only the node keys but also the subnet ranges will be updated by a new read_state call. Then although the canister_index directs canister_2 to the cache of subnet A but the final check of canister ranges will fail. 

Thanks for your explanation. It makes sense to me.

ic-agent/src/agent/mod.rs
```diff
@@ -416,14 +426,74 @@ impl Agent {
...
+    ) -> Result<Vec<u8>, AgentError> {
+        let response = if self.verify_query_signatures {
+            let (response, subnet) = futures_util::try_join!(
+                self.query_endpoint::<QueryResponse>(effective_canister_id, signed_query),
+                self.get_subnet_by_canister(&effective_canister_id)
+            )?;
+            if response.signatures().is_empty() {
+                return Err(AgentError::MissingSignature);
+            }
+            for signature in response.signatures() {
```
ShuoWangNSL, line 456: There is a timestamp in the signature for the query response.

ic-agent/src/agent/mod.rs
```diff
@@ -446,14 +459,80 @@ impl Agent {
...
+                const DER_PREFIX: [u8; 12] = [48, 42, 48, 5, 6, 3, 43, 101, 112, 3, 33, 0];
+                if node_key[..12] != DER_PREFIX {
+                    return Err(AgentError::DerPrefixMismatch {
+                        expected: DER_PREFIX.to_vec(),
+                        actual: node_key[..12].to_vec(),
+                    });
+                }
+                let pubkey =
+                    VerificationKey::try_from(<[u8; 32]>::try_from(&node_key[12..]).unwrap())
+                        .map_err(|_| AgentError::QuerySignatureVerificationFailed)?;
```
ericswanson-dfinity, line 516: Suggestion: use distinct errors (or differentiate by an enum member or something) rather than three possibilities that all end up as `QuerySignatureVerificationFailed`.  It's frustrating enough to investigate a signature verification failure when the only possibility is that the verify() method failed
```suggestion
                        .map_err(|_| AgentError::MalformedVerificationKey)?;
```

ic-agent/src/agent/mod.rs
```diff
@@ -446,14 +459,80 @@ impl Agent {
...
+                let sig = Signature::from(
+                    <[u8; 64]>::try_from(&signature.signature[..])
+                        .map_err(|_| AgentError::QuerySignatureVerificationFailed)?,
```
ericswanson-dfinity, line 519:
```suggestion
                        .map_err(|_| AgentError::MalformedSignature)?,
```

ic-agent/src/agent/mod.rs
```diff
@@ -446,14 +459,80 @@ impl Agent {
...
+                );
+
+                if pubkey.verify(&sig, &signable).is_err() {
+                    return Err(AgentError::QuerySignatureVerificationFailed);
```
ericswanson-dfinity, line 523: Is there a good reason to discard the error cause here?

ic-agent/src/agent/mod.rs
```diff
@@ -446,14 +459,80 @@ impl Agent {
...
+            let (response, subnet) = futures_util::try_join!(
+                self.query_endpoint::<QueryResponse>(effective_canister_id, signed_query),
+                self.get_subnet_by_canister(&effective_canister_id)
+            )?;
+            if response.signatures().is_empty() {
+                return Err(AgentError::MissingSignature);
+            }
+            for signature in response.signatures() {
+                if OffsetDateTime::now_utc()
+                    - OffsetDateTime::from_unix_timestamp_nanos(signature.timestamp as _).unwrap()
```
ericswanson-dfinity, line 491: Can we return an error rather than panic here?

ic-agent/src/agent/mod.rs
```diff
@@ -891,6 +976,39 @@ impl Agent {
...
+            .get_subnet_by_canister(canister);
+        if let Some(subnet) = subnet {
+            Ok(subnet)
+        } else {
+            let cert = self
+                .read_state_raw(vec![vec!["subnet".into()]], *canister)
+                .await?;
+            let time = leb128::read::unsigned(&mut lookup_value(&cert.tree, [b"time".as_ref()])?)?;
+            if (OffsetDateTime::now_utc()
+                - OffsetDateTime::from_unix_timestamp_nanos(time as _).unwrap())
```
ericswanson-dfinity, line 997: Can we return an error rather than panic here?

ic-agent/src/lib.rs
```diff
@@ -103,11 +103,12 @@
 //! generally available, additional details about the versions supported will
 //! be available here.
 
-#![deny(
+#![warn(
```
ericswanson-dfinity, line 106: Intentionally left in?

ic-agent/src/agent/mod.rs
adamspofford-dfinity: A u64 nanosecond timestamp is always within range; this call cannot panic without a breaking change in `time`.

ic-agent/src/lib.rs
adamspofford-dfinity: Yes. I had been meaning to remove it, since it makes development more annoying and CI still covers it with `-Dwarnings`.

ic-agent/src/agent/mod.rs
adamspofford-dfinity: The crate surfaces no underlying information about the failure, to prevent side-channels. The only other kinds of errors than a generic 'verification failed' would have happened in the `try_from`s.

ic-transport-types/src/lib.rs
```diff
@@ -122,9 +125,75 @@ pub enum QueryResponse {
...
+impl QueryResponse {
+    /// Returns the signable form of the query response, as described in
+    /// [the spec](https://internetcomputer.org/docs/current/references/ic-interface-spec#http-query).
+    /// This is what is signed in the `signatures` fields.
+    pub fn signable(&self, request_id: RequestId, timestamp: u64) -> Vec<u8> {
+        #[derive(Serialize)]
+        #[serde(tag = "status", rename_all = "snake_case")]
+        enum QueryResponseSignable<'a> {
+            Replied {
+                reply: RequestId, // polyfill until hash_of_map is figured out
```
ShuoWangNSL, line 154: All subnets including NNS has upgraded to support nested maps. This can be changed accrodingly.

ic-agent/src/agent/mod.rs
adamspofford-dfinity: See other comment.

icx/src/main.rs
```diff
@@ -399,6 +399,7 @@ async fn main() -> Result<()> {
                         result.unwrap_or(Err(AgentError::TimeoutWaitingForResponse()))
                     }
                     SubCommand::Query(_) => {
```
ShuoWangNSL, line 401: FYI. I find some typos in error messages in this file when I am working on a demo.
`Got an error when make` should be `Got an error when making` .
`Got an error when send` should be `Got an error when sending` .

adamspofford-dfinity: I've been meaning to clean up icx for a while, there's lots that's wrong in here besides that. It'll be a separate PR though.

ericswanson-dfinity (approved):

PR #481, opened 2023-10-10
feat: Round `ingress_expiry` to the minute

adamspofford-dfinity:
If the agent's configured ingress expiry duration is >1m, this rounds the ingress expiry *timestamp* to the closest minute (and otherwise, rounds down to the second). This will hopefully reduce cache misses, since a nanosecond difference in the ingress expiry timestamp means a different request ID and is thus uncacheable.

Implements SDK-1259.

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -1285,7 +1285,7 @@ impl<'agent> UpdateBuilder<'agent> {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "reqwest"))]
 mod offline_tests {
     use super::*;
     // Any tests that involve the network should go in agent_test, not here.
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -331,19 +331,13 @@ impl Agent {
 
     fn get_expiry_date(&self) -> u64 {
         let expiry_raw = OffsetDateTime::now_utc() + self.ingress_expiry;
-        let rounded = if self.ingress_expiry.as_secs() > 60 {
-            let mut rounded_minute = expiry_raw
-                .replace_second(0)
-                .unwrap()
-                .replace_nanosecond(0)
-                .unwrap();
+        let mut rounded = expiry_raw.replace_nanosecond(0).unwrap();
+        if self.ingress_expiry.as_secs() > 60 {
+            rounded = rounded.replace_second(0).unwrap();
             if expiry_raw.second() >= 30 {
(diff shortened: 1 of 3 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -330,12 +330,21 @@ impl Agent {
...
+                .unwrap()
+                .replace_nanosecond(0)
+                .unwrap();
+            if expiry_raw.second() >= 30 {
+                rounded_minute += Duration::from_secs(60);
+            }
+            rounded_minute
+        } else {
+            expiry_raw.replace_nanosecond(0).unwrap()
+        };
```
ericswanson-dfinity, line 346: It seems like we're unconditionally doing `.replace_nanosecond(0)`.  Could we do it in one place?

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

ic-agent/src/agent/mod.rs
```diff
@@ -1275,3 +1278,31 @@ impl<'agent> UpdateBuilder<'agent> {
...
+
+    #[test]
+    fn rounded_expiry() {
+        let agent = Agent::builder()
+            .with_url("http://not-a-real-url")
+            .build()
+            .unwrap();
+        let mut prev_expiry = None;
+        let mut num_timestamps = 0;
+        for _ in 0..6 {
```
roelstorms, line 1295: Just out of curiosity, why aren't we using the iterator from the for loop instead of num_timestamps?

ic-agent/src/agent/mod.rs
```diff
@@ -330,12 +330,15 @@ impl Agent {
...
     fn get_expiry_date(&self) -> u64 {
-        // TODO(hansl): evaluate if we need this on the agent side (my hunch is we don't).
-        let permitted_drift = Duration::from_secs(60);
-        self.ingress_expiry
-            .as_nanos()
-            .saturating_add(OffsetDateTime::now_utc().unix_timestamp_nanos() as u128)
-            .saturating_sub(permitted_drift.as_nanos()) as u64
+        let expiry_raw = OffsetDateTime::now_utc() + self.ingress_expiry;
+        let mut rounded = expiry_raw.replace_nanosecond(0).unwrap();
+        if self.ingress_expiry.as_secs() > 60 {
```
roelstorms, line 335: Why is rounding only done if ingress_expiry is larger than 60 seconds? 
Is this aligned between the BN team (caching), agent-js and agent-rs?

PR #482, opened 2023-10-30
chore: make Envelope struct public

massimoalbarello:
# Description

For the [IC WebSocket Gateway](https://github.com/omnia-network/ic-websocket-gateway) [we need](https://github.com/omnia-network/ic-websocket-gateway/blob/3cc3499e21eb8314f499a29bcc55a79fe359c3c9/src/ic-websocket-gateway/src/client_connection_handler.rs#L37) the `Envelope` struct to be public also outside the `ic-agent` crate as the gateway needs to deserialize messages coming from clients and containing a serialized envelope.

# How Has This Been Tested?
Only made `Envelope` struct public.

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -12,7 +12,7 @@ pub use agent_error::AgentError;
 pub use builder::AgentBuilder;
 #[doc(inline)]
 pub use ic_transport_types::{
-    signed, EnvelopeContent, RejectCode, RejectResponse, ReplyResponse, RequestStatusResponse,
+    signed, Envelope, EnvelopeContent, RejectCode, RejectResponse, ReplyResponse, RequestStatusResponse,
 };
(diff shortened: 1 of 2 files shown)


sesi200 (approved):

ic-agent/src/agent/mod.rs
```diff
@@ -12,7 +12,7 @@ pub use agent_error::AgentError;
 pub use builder::AgentBuilder;
 #[doc(inline)]
 pub use ic_transport_types::{
-    signed, EnvelopeContent, RejectCode, RejectResponse, ReplyResponse, RequestStatusResponse,
+    signed, Envelope, EnvelopeContent, RejectCode, RejectResponse, ReplyResponse, RequestStatusResponse,
```
sesi200, line 15:
```suggestion
    signed, Envelope, EnvelopeContent, RejectCode, RejectResponse, ReplyResponse,
    RequestStatusResponse,
```
rustfmt complains about this

PR #484, opened 2023-11-01
feat: extend agent with a route provider feature

nikolay-komarevskiy:
# Description

_API Boundary Node (BN) Discovery Library_  (which is in a very recent [development](https://gitlab.com/dfinity-lab/public/ic/-/tree/master/rs/boundary_node/discower_bowndary?ref_type=heads)) is the prerequisite for the [IC-720: BN Decentralization Feature](https://forum.dfinity.org/t/boundary-node-roadmap/15562). This library will have two main responsibilities:
1. **Discovery of the API Boundary Nodes (domains)**.  Users (clients) of the IC need to know, which domains to use when "talking" to the IC. Currently, there is only one domain `ic0.app` owned by _Dfinity_. However, once BNs are decentralized, each BN will have it's own/unique domain. Moreover, BNs and their domain records could be added (or removed) from the IC registry via nns proposals. Thus, clients would need an api to discover, which domains currently belong to the IC ecosystem. _Note_: the exact BN domain discovery mechanism is still being discussed.  This could either be accomplished via [read_state](https://internetcomputer.org/docs/current/references/ic-interface-spec#http-read-state) calls or by fetching records from the registry canister.
3. **Routing - dynamically providing url (BN domain) on demand**

**Future usages of the library**:
- browsers
- http gateway
- icx proxy

The library is being developed within the [ic](https://gitlab.com/dfinity-lab/public/ic/-/tree/master/rs/boundary_node/discower_bowndary?ref_type=heads) repo and it uses `ic-agent` as a dependency. However, in the near future, `ic-agent` could/should include some functionalities of the library or even library as a whole.

This pull request kicks-off the first migration/integration step of the library with the `ic-agent`.
Namely,  `ic-agent` is extended with a custom `RouteProvider` feature.

# How Has This Been Tested?

- Existing tests should fully cover the e2e functionality of the `RoundRobinRouteProvider` which is now used as default route provider for the `agent`.
- In addition, several unit tests were added for the `RoundRobinRouteProvider`.

# Checklist:

### DIFF
--- a/ic-agent/src/agent/http_transport/hyper_transport.rs
+++ b/ic-agent/src/agent/http_transport/hyper_transport.rs
@@ -3,21 +3,17 @@ pub use hyper;
 
 use std::{any, error::Error, future::Future, marker::PhantomData, sync::atomic::AtomicPtr};
 
-use http::uri::{Authority, PathAndQuery};
 use http_body::{LengthLimitError, Limited};
 use hyper::{
-    body::{Bytes, HttpBody},
-    client::HttpConnector,
-    header::CONTENT_TYPE,
-    service::Service,
-    Client, Method, Request, Response, Uri,
+    body::HttpBody, client::HttpConnector, header::CONTENT_TYPE, service::Service, Client, Method,
+    Request, Response,
 };
 use hyper_rustls::{HttpsConnector, HttpsConnectorBuilder};
 
 use crate::{
     agent::{
         agent_error::HttpErrorPayload,
-        http_transport::{IC0_DOMAIN, IC0_SUB_DOMAIN},
+        http_transport::route_provider::{RoundRobinRouteProvider, RouteProvider},
         AgentFuture, Transport,
     },
     export::Principal,
@@ -28,7 +24,7 @@ use crate::{
 #[derive(Debug)]
 pub struct HyperTransport<B1, S = Client<HttpsConnector<HttpConnector>, B1>> {
     _marker: PhantomData<AtomicPtr<B1>>,
-    url: Uri,
+    route_provider: Box<dyn RouteProvider>,
     max_response_body_size: Option<usize>,
     service: S,
 }
@@ -87,7 +83,7 @@ where
 
 impl<B1: HyperBody> HyperTransport<B1> {
     /// Creates a replica transport from a HTTP URL.
-    pub fn create<U: Into<Uri>>(url: U) -> Result<Self, AgentError> {
+    pub fn create<U: Into<String>>(url: U) -> Result<Self, AgentError> {
         let connector = HttpsConnectorBuilder::new()
             .with_webpki_roots()
             .https_or_http()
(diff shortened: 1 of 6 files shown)


adamspofford-dfinity (commented):
I am hesitant to introduce yet another context-free trait object into the agent. Is there any reason a user would configure anything other than a round-robin strategy for a particular list of URLs? If not, this produces much less complexity with the user just providing a list of URLs, and eliminating the trait.

nikolay-komarevskiy:
> I am hesitant to introduce yet another context-free trait object into the agent. Is there any reason a user would configure anything other than a round-robin strategy for a particular list of URLs? If not, this produces much less complexity with the user just providing a list of URLs, and eliminating the trait.

@adamspofford-dfinity I understand the concern. Imho, `RouteProvider` should allow for multiple implementations due to the following reasons. Route providers behavior might be configurable based on: latency, geo-location. Also periodic refreshing of the routes list (domains) from the registry could take place, as some API BNs can be added/removed. For round-robin and route providers, it might also make sense to evict a route, if it fails frequently. Does this answer the question?

ic-agent/src/agent/http_transport/reqwest_transport.rs
```diff
@@ -7,29 +7,28 @@ pub use reqwest;
...
-        http_transport::{IC0_DOMAIN, IC0_SUB_DOMAIN},
+        http_transport::route_provider::{RoundRobinRouteProvider, RouteProvider},
         AgentFuture, Transport,
     },
     export::Principal,
     AgentError, RequestId,
 };
 
 /// A [`Transport`] using [`reqwest`] to make HTTP calls to the Internet Computer.
-#[derive(Debug)]
```
adamspofford-dfinity, line 24: was this intentional?

ic-agent/src/agent/http_transport/reqwest_transport.rs
```diff
@@ -7,29 +7,28 @@ pub use reqwest;
...
 pub struct ReqwestTransport {
-    url: Url,
+    route_provider: Box<dyn RouteProvider>,
     client: Client,
     max_response_body_size: Option<usize>,
 }
 
 #[doc(hidden)]
-pub use ReqwestTransport as ReqwestHttpReplicaV2Transport; // deprecate after 0.24
```
adamspofford-dfinity, line 32: This should be left in.

ic-agent/src/agent/http_transport/hyper_transport.rs
```diff
@@ -285,32 +263,38 @@ where
...
-        test("https://foo.ic1.app", "https://foo.ic1.app/api/v2");
-        test("https://ic0.app.ic1.app", "https://ic0.app.ic1.app/api/v2");
+        test("https://ic1.app", "https://ic1.app/api/v2/");
+        test("https://foo.ic1.app", "https://foo.ic1.app/api/v2/");
+        test("https://ic0.app.ic1.app", "https://ic0.app.ic1.app/api/v2/");
 
-        test("https://fooic0.app", "https://fooic0.app/api/v2");
-        test("https://fooic0.app.ic0.app", "https://ic0.app/api/v2");
+        test("https://fooic0.app", "https://fooic0.app/api/v2/");
+        test("https://fooic0.app.ic0.app", "https://ic0.app/api/v2/");
```
adamspofford-dfinity, line 298: There should also be tests against icp0.io and icp-api.io, as well as at least one principal instead of only subdomains without dashes.

ic-agent/src/agent/http_transport/route_provider.rs
```diff
@@ -0,0 +1,96 @@
...
+    AgentError,
+};
+
+/// A [`RouteProvider`] for dynamic generation of routing urls.
+pub trait RouteProvider: std::fmt::Debug + Send + Sync {
+    /// Generate next routing url
+    fn route(&self) -> Result<Url, AgentError>;
+}
+
+/// A simple implementation of the [`RouteProvider``] which produces an even distribution of the urls from the input ones.
```
adamspofford-dfinity, line 19:
```suggestion
/// A simple implementation of the [`RouteProvider`] which produces an even distribution of the urls from the input ones.
```

ic-agent/src/agent/http_transport/reqwest_transport.rs
nikolay-komarevskiy: accidental, thanks for spotting, changed back

ic-agent/src/agent/http_transport/hyper_transport.rs
nikolay-komarevskiy: Not sure i fully got it. Could You write the assertions? I will take care of the implementation if tests fail.

ic-agent/src/agent/http_transport/reqwest_transport.rs
nikolay-komarevskiy: same here, deleted accidentally

adamspofford-dfinity (approved):

adamspofford-dfinity (commented):
Can you also add an entry to the changelog? Also I mistyped the localhost one, it should be http not https.

PR #490, opened 2023-11-15
feat: implement chunked wasm API

adamspofford-dfinity:
This implements the large/chunked wasm feature in the management canister, which permits uploading chunks of a wasm module one at a time, and then making a final call to assemble the module out of the chunks. It additionally includes a generic install builder that will switch between install_code and install_chunked_code depending on module size, which the install_code builder cannot do due to providing an AsyncCall impl.


ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -447,10 +462,8 @@ impl<'agent, 'canister: 'agent> InstallCodeBuilder<'agent, 'canister> {
     }
     /// Set the argument with multiple arguments as tuple to the installation,
     /// which will be passed to the init method of the canister. Can be called at most once.
-    pub fn with_args(mut self, tuple: impl candid::utils::ArgumentEncoder) -> Self {
-        if self.arg.0.is_some() {
-            panic!("argument is being set more than once");
-        }
+    pub fn with_args(mut self, tuple: impl ArgumentEncoder) -> Self {
+        assert!(self.arg.0.is_none(), "argument is being set more than once");
```
ericswanson-dfinity, line 466: If I understand this, this means:
- in a debug build, calling with_args more than once will result in a panic
- in a release build, calling with_args more than once will silent replace the args

Is that how it is, and desired?

adamspofford-dfinity: No. That is `debug_assert!`. The replacement is exactly equivalent, just more idiomatic.

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -507,6 +520,275 @@ impl<'agent, 'canister: 'agent> AsyncCall<()> for InstallCodeBuilder<'agent, 'ca
...
+    /// Invoke the installation process. The returned stream must be awaited to completion; it is used to track progress,
+    /// as installation may take arbitrarily long, and is intended to be passed to functions like `indicatif::ProgressBar::wrap_stream`.
+    /// There are exactly [`size_hint().0`](Stream::size_hint) steps.
+    pub fn call_and_wait_with_progress(
+        self,
+    ) -> impl Stream<Item = Result<(), AgentError>> + 'builder {
+        let stream_res = /* try { */ (move || {
+            let arg = self.arg.serialize()?;
+            let stream: BoxStream<'_, _> =
+                if self.wasm.len() + arg.len() < (1.85 * 1024. * 1024.) as usize {
```
ericswanson-dfinity, line 724: where does 1.85 come from?

adamspofford-dfinity: It's a little bit under 2, to account for the rest of the message.

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -507,6 +521,280 @@ impl<'agent, 'canister: 'agent> AsyncCall<()> for InstallCodeBuilder<'agent, 'ca
...
+            .await
+    }
+
+    /// Invoke the installation process. The returned stream must be awaited to completion; it is used to track progress,
+    /// as installation may take arbitrarily long, and is intended to be passed to functions like `indicatif::ProgressBar::wrap_stream`.
+    /// There are exactly [`size_hint().0`](Stream::size_hint) steps.
+    pub fn call_and_wait_with_progress(
+        self,
+    ) -> impl Stream<Item = Result<(), AgentError>> + 'builder {
+        let stream_res = /* try { */ (move || {
```
ericswanson-dfinity, line 726: what is going on with this?
> /* try { */

adamspofford-dfinity: An immediately-invoked closure is being used to emulate a try block (local scope for `?`), which does not yet exist.

ericswanson-dfinity (approved):

PR #494, opened 2023-11-17
chore!: bump candid

chenyan-dfinity:
# Description

Bump Candid to 0.10. Downstream libraries need to bump Candid to 0.10 as well.

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.30.3"
+version = "0.30.2"
 authors = ["DFINITY Stiftung <sdk@dfinity.org>"]
 edition = "2021"
 repository = "https://github.com/dfinity/agent-rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.30.2"
+version = "0.30.3"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -10,7 +10,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.30.2"
+version = "0.30.3"
```
adamspofford-dfinity, line 13: We don't need to make a release just for this, do we?

chenyan-dfinity: Would be good to have a release, so that we can bump the downstream candid version.

adamspofford-dfinity: Well, it is a breaking change, so it cannot be a point release as you have marked it. I think it would be better to leave off the release for now, and then we can decide later whether a full version just containing the candid bump is necessary.

chenyan-dfinity: Okay, sounds good.

adamspofford-dfinity (approved):

ericswanson-dfinity (approved):

Cargo.toml
frederikrothenberger: FYI, I'd like to have a release of the agent in order to upgrade the `ic` repo to candid `0.10` so that could finally make use of that in Internet Identity.

PR #498, opened 2023-12-01
feat: Misc icx fixes

adamspofford-dfinity:
The primary target of this PR is updating the URLs that the root key is fetched for to include icp-api.io and icp0.io, while making the default port 4943, but I also took the opportunity to fix some documentation and panics, add a `--ic` flag as shorthand for mainnet, and make the flags repositionable.


icx/src/main.rs
```diff
@@ -251,16 +263,28 @@ fn print_idl_blob(
...
 
-async fn fetch_root_key_from_non_ic(agent: &Agent, replica: &str) -> Result<()> {
-    let normalized_replica = replica.strip_suffix('/').unwrap_or(replica);
-    if normalized_replica != DEFAULT_IC_GATEWAY {
-        agent
-            .fetch_root_key()
-            .await
-            .context("Failed to fetch root key from replica")?;
+async fn fetch_root_key_from_non_ic(agent: &Agent, replica: &Url) -> Result<()> {
+    if is_mainnet(replica) {
```
ericswanson-dfinity, line 267: Shouldn't this logic be the opposite?  We should never fetch the root key on mainnet, and should always fetch it elsewhere.
```suggestion
    if !is_mainnet(replica) {
```

ericswanson-dfinity (changes requested):

icx/src/main.rs
ericswanson-dfinity: We could also consider the approach that icx-proxy takes: adding a `--fetch-root-key` parameter.

https://github.com/dfinity/ic/blob/master/rs/boundary_node/icx_proxy/src/main.rs#L83

PR #509, opened 2024-02-05
feat: add fetch_api_boundary_nodes API to ic-agent

nikolay-komarevskiy:
# Description

API boundary nodes were recently added to the certified state tree. Currently, there are no API boundary nodes and the subtree is empty. accordingly. API boundary nodes will be added to the registry via NNS proposals at some later stage.

IC clients should be able to retrieve all these API BNs in a convenient way.

This PR extends `ic-agent` API with a new method:
`pub async fn fetch_api_boundary_nodes() -> Vec<ApiBoundaryNode>`

# How Has This Been Tested?

1. Run this [api_boundary_node_decentralization_test](https://sourcegraph.com/github.com/dfinity/ic@db9595a8be755949a28e0af74c6bae85fa1a9c3d/-/blob/rs/tests/boundary_nodes/BUILD.bazel?L142) `system-test`. This test converts two unassigned nodes into API BNs.
2. Instantiate an `agent` with any IC replica url provided in `system-test` logs during test execution.
3. Retrieve API BNs via `let api_bns = agent.fetch_api_boundary_nodes(effective_canister_id).await`.
4. Verify that two  API BNs are discovered within the `State Tree`.

# Checklist:


ic-agent/src/agent/mod.rs
```diff
@@ -1019,6 +1021,21 @@ impl Agent {
             }
         }
     }
+
+    /// Retrieve all existing API Boundary Nodes from the State Tree.
+    pub async fn fetch_api_boundary_nodes(
+        &self,
+        effective_canister_id: Principal,
+    ) -> Result<Arc<Vec<ApiBoundaryNode>>, AgentError> {
```
adamspofford-dfinity, line 1029: This should just be Vec<ApiBoundaryNode>. If the user wants it in an Arc they can place it there themselves.

ic-agent/src/agent/mod.rs
```diff
@@ -1328,6 +1345,17 @@ pub(crate) struct Subnet {
     canister_ranges: RangeInclusiveSet<Principal, PrincipalStep>,
 }
 
+/// API Boundary Node, which routes /api calls to IC replica nodes.
+#[derive(Debug)]
+pub struct ApiBoundaryNode {
+    /// Domain name
+    _domain: String,
```
adamspofford-dfinity, line 1352: Presumably these are underscored to silence an 'unused' warning? They would also be silenced by marking the fields `pub`, which I think you have forgotten to do as users cannot do anything with `ApiBoundaryNode`.

adamspofford-dfinity (approved):

ic-agent/src/agent/mod.rs
nikolay-komarevskiy: > Presumably these are underscored to silence an 'unused' warning?

Yes, exactly, changed to `pub`.

PR #511, opened 2024-02-06
feat: add pre signed read state call wait

NikolasHai:
This MR proposes the following changes:

Currently, the agent can only be used with a private key that is known to the agent. If someone wants to sign their request outside of the agent, they cannot forward this signed request using the agent. 

The agent already offers a method that lets a user forward a signed `Envelope` of `EnvelopeContent::Call` to the IC. This MR adds the same functionality for a signed `Envelope` of EnvelopeContent::ReadState to the IC. 

The method that is added is similar to the already existing `wait(..)` method but also takes in the signed request.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1822,6 +1822,7 @@ version = "0.0.0"
 dependencies = [
  "candid",
  "ic-agent",
+ "ic-certification",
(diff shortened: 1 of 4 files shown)


adamspofford-dfinity (approved):
LGTM

PR #512, opened 2024-02-07
fix: Change stored_chunks return type

adamspofford-dfinity:
`stored_chunks` is destined to change from returning `blob` to `record { hash: blob }`. This changes the Rust type to the latter, with a manual deserialization impl that also accepts the former.


ericswanson-dfinity (changes requested):
changelog

ericswanson-dfinity (approved):

PR #518, opened 2024-02-21
chore: Use dfinity/setup-dfx for e2e

adamspofford-dfinity:


.github/workflows/e2e.yml
```diff
@@ -34,6 +34,10 @@ jobs:
         run: bash .github/workflows/provision-linux.sh
         env:
           INSTALL_DFX_VERSION: ${{ matrix.dfx }}
+      - name: Install DFX
+        uses: dfinity/setup-dfx@main
+        with:
+          dfx-version: "0.16.1"
```
ericswanson-dfinity, line 40:
```suggestion
          dfx-version: ${{ matrix.dfx }}
```

ericswanson-dfinity (changes requested):

ericswanson-dfinity (approved):

adamspofford-dfinity (commented):
@lwshang I've noticed that for a while but we could add some eventually, and then we'd feel silly for having gotten rid of the setup. To be revisited later, maybe.

PR #521, opened 2024-02-22
fix: check read_state timestamps

adamspofford-dfinity:
This expands the timestamp check used for subnet keys to any certificate verification. 

The tests were also updated to use 4943 as the default port, and to be compatible with being run twice against the same replica.

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -882,7 +882,6 @@ impl Agent {
         let time = lookup_time(cert)?;
         if (OffsetDateTime::now_utc()
             - OffsetDateTime::from_unix_timestamp_nanos(time.into()).unwrap())
-        .abs()
             > self.ingress_expiry
         {
             Err(AgentError::CertificateOutdated(self.ingress_expiry))
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -884,6 +884,7 @@ impl Agent {
         let time = lookup_time(cert)?;
         if (OffsetDateTime::now_utc()
             - OffsetDateTime::from_unix_timestamp_nanos(time.into()).unwrap())
+        .abs()
             > self.ingress_expiry
         {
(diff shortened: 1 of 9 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -877,6 +878,19 @@ impl Agent {
             .map_err(|_| AgentError::CertificateVerificationFailed())
     }
 
+    fn verify_cert_timestamp(&self, cert: &Certificate) -> Result<(), AgentError> {
+        let time = lookup_time(cert)?;
+        if (OffsetDateTime::now_utc()
+            - OffsetDateTime::from_unix_timestamp_nanos(time.into()).unwrap())
+        .abs()
```
ericswanson-dfinity, line 885: Maybe this should check for timestamps that are in the future, but calling abs() and returning CertificateOutdated doesn't seem correct.

ericswanson-dfinity: That sounds like incorrect behavior on the part of agent-js then.  We shouldn't copy it here.

ericswanson-dfinity (changes requested):

ic-agent/src/agent/mod.rs
ericswanson-dfinity: See PERMITTED_DRIFT for how this is handled elsewhere https://github.com/dfinity/ic/blob/master/rs/constants/src/lib.rs#L15

adamspofford-dfinity: AIUI `PERMITTED_DRIFT` is for internal replica drift, where one subnet drifts from the other, and that this doesn't account for the caller's drift; is this accurate?

ericswanson-dfinity: I do not know the answer to that.

adamspofford-dfinity: Well, for now I'll just delete the abs check then. If you're that far in the past, the replica would have already rejected your expiry timeout as being in the past.

ericswanson-dfinity (approved):

PR #526, opened 2024-03-07
fix: fetch_root_key refuses to play along if called on the mainnet

ericswanson-dfinity:
# Description

It's a mistake to call `Agent.fetch_root_key()` on the mainnet.  This makes it also an error, and also zeroes out the root key so that the agent would fail to verify any signatures even if the caller ignored the error. 

# How Has This Been Tested?

Added a test

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added node metrics, ECDSA, and Bitcoin functions to `MgmtMethod`. Most do not have wrappers in `ManagementCanister` because only canisters can call these functions.
 * Added `FetchCanisterLogs` function to `MgmtMethod` and a corresponding wrapper to `ManagementCanister`.
 * Updated the `ring` crate to 0.17.7.  `ring` 0.16 has a bug where it requires incorrect Ed25519 PEM encoding. 0.17.7 fixes that and is backwards compatible.
+* Agent::fetch_root_key() now returns an error, and sets its root key to an empty vector, if called on the mainnet.
 
 ## [0.33.0] - 2024-02-08
 
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -330,6 +330,11 @@ impl Agent {
         }
         let status = self.status().await?;
         let root_key = match status.root_key {
+            Some(key) if key[..] == IC_ROOT_KEY[..] => {
+                // even if the caller ignores this error, we're done here.
+                self.set_root_key(vec![]);
+                return Err(AgentError::NeverFetchRootKeyOnMainNet());
+            }
             Some(key) => key,
             None => return Err(AgentError::NoRootKeyInStatus(status)),
         };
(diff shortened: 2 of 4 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -330,6 +330,11 @@ impl Agent {
         }
         let status = self.status().await?;
         let root_key = match status.root_key {
+            Some(key) if key[..] == IC_ROOT_KEY[..] => {
```
mraszyk, line 333: This wouldn't apply in case of an actual man-in-the-middle-attack.

ericswanson-dfinity: Correct. What this is meant to protect against is accidentally fetching the root key on mainnet, which opens the door to the man-in-the-middle attack.

This applies in the case where you are erroneously fetching the root key on mainnet.  Prior to this change you won't notice, because everything works: you fetch the root key from mainnet, replace your root key with it (the same key), and go about your day.  You shouldn't have, but no harm done, right? It "works" locally, it "works" on mainnet, it "works" on any testnet.  It "works" everywhere; there's just one problem...

Then one day, you connect with the agent, only there's been a DNS hack and you connect somewhere else: a man-in-the-middle.  You fetch the root key, like you always do.  It's not the the IC_ROOT_KEY this time, but you don't notice, and all the signatures check out. You're fallen for the man-in-the-middle attack.

This change is intended to make you notice that you are fetching the root key from mainnet (opening yourself up to this attack) and stop doing that.

Does that make sense?

mraszyk: It does make sense to _nudge_ people not to fetch the root key from mainnet. But if we wanted to _make sure_ that people don't fetch the root key from mainnet, we'd need to take the URL into account.

ericswanson-dfinity: There are [other places](https://github.com/dfinity/sdk/blob/master/src/dfx-core/src/config/model/network_descriptor.rs#L83) that look at the URL and later [skip the call](https://github.com/dfinity/sdk/blob/master/src/dfx-core/src/network/root_key.rs#L10) or [flag this as an error](https://github.com/dfinity/sdk/blob/master/src/dfx-core/src/network/root_key.rs#L29). But I wouldn't consider them as _making sure_, because there are a lot of ways to express the same URL.

This method however does seem like a way of making sure that you aren't blithely calling fetch_root_key against , because if you do, the only time it doesn't notice is if there is a MITM attack going on.

sesi200 (approved):

ic-agent/src/agent/mod.rs
mraszyk: Sounds good. Thank you for your explanations!

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added node metrics, ECDSA, and Bitcoin functions to `MgmtMethod`. Most do not have wrappers in `ManagementCanister` because only canisters can call these functions.
 * Added `FetchCanisterLogs` function to `MgmtMethod` and a corresponding wrapper to `ManagementCanister`.
 * Updated the `ring` crate to 0.17.7.  `ring` 0.16 has a bug where it requires incorrect Ed25519 PEM encoding. 0.17.7 fixes that and is backwards compatible.
+* Agent::fetch_root_key() now returns an error, and sets its root key to an empty vector, if called on the mainnet.
```
lwshang, line 16:
```suggestion
* Agent::fetch_root_key() now returns an error, and sets its root key to an empty vector, if got the same root key as the mainnet. This reduce the potentiality of MITM attack.
```

PR #528, opened 2024-03-09
chore: Check features individually

adamspofford-dfinity:
This PR changes the cargo clippy setup to use [`cargo-hack`](https://github.com/taiki-e/cargo-hack), to check each feature individually instead of only the default features or only the full set of features. Two commands are used so that features enabled in dev-dependencies don't leak into checking normal builds.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,10 +21,17 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hack@0.6.21
+
       - name: Run Lint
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -21,10 +21,17 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hack@0.6.21
```
lwshang, line 26: Do we need this specified version? What about making the action always install latest `cargo -hack`?

adamspofford-dfinity: Changes in the interface or behavior could break the build or stop checking a particular case; if this happens, I would prefer it not happen without warning.

PR #531, opened 2024-03-12
feat!: change ReplicaError to Certified/UncertifiedReject

adamspofford-dfinity:


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+* Changed `AgentError::ReplicaError` to `CertifiedReject` or `UncertifiedReject`. `CertifiedReject`s went through consensus, and `UncertifiedReject`s did not.
```
ericswanson-dfinity, line 11: Since this is a breaking change that will cause compile errors (`ReplicaError` is not defined), is there something we can write here and/or in AgentError about what a developer should do to update their code?

adamspofford-dfinity: This is more or less the point of the changelog. I don't think there is an agent-rs spot in the portal, other than the page that just mentions it exists.
Unfortunately I don't think there's any way to add something that creates a useful message. If it were a normal symbol, you can cheat with `const X: () = { panic!("see Y") }`, but since this is used as a *pattern*, I don't think there's an equivalent.

ericswanson-dfinity (approved):

CHANGELOG.md
adamspofford-dfinity: Unless you're saying to be more descriptive in the changelog as one solution.

PR #538, opened 2024-03-19
fix: lookup path is absent

sesi200:
# Description

Occasionally, various operations in dfx fail with `Lookup path is absent`. Current best guess: After submitting a request the request id doesn't always appear quickly enough in the state tree. This should not result in an error because there is no guarantee that the request id appears in the state tree within a very short time

(Hopefully) fixes [SDK-1396](https://dfinity.atlassian.net/browse/SDK-1396)

# How Has This Been Tested?

TBD

# Checklist:

[SDK-1396]: https://dfinity.atlassian.net/browse/SDK-1396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -1033,9 +1033,11 @@ impl Agent {
         let paths: Vec<Vec<Label>> =
             vec![vec!["request_status".into(), request_id.to_vec().into()]];
 
-        let cert = self.read_state_raw(paths, effective_canister_id).await?;
-
-        lookup_request_status(cert, request_id)
+        match self.read_state_raw(paths, effective_canister_id).await {
+            Ok(cert) => lookup_request_status(cert, request_id),
+            Err(AgentError::LookupPathAbsent(_)) => Ok(RequestStatusResponse::Unknown),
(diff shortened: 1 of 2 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -1033,9 +1033,11 @@ impl Agent {
         let paths: Vec<Vec<Label>> =
             vec![vec!["request_status".into(), request_id.to_vec().into()]];
 
-        let cert = self.read_state_raw(paths, effective_canister_id).await?;
-
-        lookup_request_status(cert, request_id)
+        match self.read_state_raw(paths, effective_canister_id).await {
+            Ok(cert) => lookup_request_status(cert, request_id),
+            Err(AgentError::LookupPathAbsent(_)) => Ok(RequestStatusResponse::Unknown),
```
mraszyk, line 1038: How can you get this error from `self.read_state_raw`? In other words, I don't think this line is actually reachable code.

PR #540, opened 2024-03-21
feat: add call_with_verification and call_without_verification methods to QueryBuilder

lwshang:
# Description

* Added `QueryBuilder::call_with_verification()` and `QueryBuilder::call_without_verification()` which always/never verify query signatures
  regardless the Agent level configuration from `AgentBuilder::with_verify_query_signatures`.

Part of SDK-1505

# How Has This Been Tested?

The [sdk PR](https://github.com/dfinity/sdk/pull/3668) is using the new method `call_without_verification()` which was tested in the sdk e2e tests.

# Checklist:

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -423,6 +423,7 @@ impl Agent {
 
     /// The simplest way to do a query call; sends a byte array and will return a byte vector.
     /// The encoding is left as an exercise to the user.
+    #[allow(clippy::too_many_arguments)]
     async fn query_raw(
         &self,
         canister_id: Principal,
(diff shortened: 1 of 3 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -1554,6 +1559,43 @@ impl<'agent> QueryBuilder<'agent> {
...
+                Some(true),
+            )
+            .await
+    }
+
+    /// Make a query call without signature verification. This will return a byte vector.
+    ///
+    /// Compared with [call][Self::call], this method will **never** verify the signature of the query response
+    /// regardless the Agent level configuration from [AgentBuilder::with_verify_query_signatures].
+    pub async fn call_without_verification(self) -> Result<Vec<u8>, AgentError> {
```
adamspofford-dfinity, line 1589: Why are these additional build methods instead of an additional field to the builder?

lwshang: The query is conceptually the same. The difference is how do we process the response.

For similarity, `UpdateBuilder` has `call()` and `call_and_wait()`.

PR #542, opened 2024-03-25
fix: Remove `DestinationInvalid` check from `WalletCanister::from_canister`

adamspofford-dfinity:
New replica versions report it as CanisterError instead; removing the check entirely is easier than doing this again in the future.


ericswanson-dfinity (changes requested):
changelog

ericswanson-dfinity (approved):

PR #549, opened 2024-04-24
feat: client-side ratelimiting and auto-retrying 429s

adamspofford-dfinity:
ManagementCanister::install currently does not work for very large modules due to hitting 429 after the first fifty chunks. This PR fixes that in two ways: limiting concurrent requests to 50 via semaphore, and retrying any 429s that slip through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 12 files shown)


lwshang:
Can we add some tests?

I'm aware that it is not easy to test in this case. Because the 429 http status code came from the boundary nodes. We have to test against the mainnet or some testnets instead of a local replica. It is hard. But I still insist that tests are necessary.

lwshang (approved):

PR #568, opened 2024-06-25
feat(BOUN-1168): add dynamic route provider (Discovery Library)

nikolay-komarevskiy:
# Description

API boundary nodes (BNs) [have already been](https://dashboard.internetcomputer.org/nodes?type=ApiBoundary) introduced into the Internet Computer (IC) landscape. Currently, the `agent` does not support dynamic discovery of existing API BNs or the dynamic routing of requests to these nodes based on specific strategies, such as latency. This PR enhances the `agent` by adding functionality for the dynamic routing to API BNs. In a nutshell the newly introduced [DynamicRouteProvider](https://github.com/dfinity/agent-rs/pull/568/files#diff-c3cffa98e93a87951c04778073900f72c8e3b8225b33b30a9ed2cc88545bb3c8R51)  is the implementation of the [RouteProvider trait](https://github.com/dfinity/agent-rs/blob/be929fd7967249c879f48f2f494cbfc5805a7d98/ic-agent/src/agent/http_transport/route_provider.rs#L17).

Example usage 1 - ergonomic:
```sh
#[tokio::main]
async fn main() {
    let client = Client::builder().build().expect("failed to build a client");

    // Initialize the agent with a custom dynamic transport layer
    // NOTE: this requires an async context
    let agent = Agent::builder()
        .with_discovery_transport(client)
        .await
        .build()
        .expect("failed to create an agent");
    //
    // ... use agent
    //

}
```

Example usage 2 - fully customized:
```sh
#[tokio::main]
async fn main() {
    let client = Client::builder().build().expect("failed to build a client");
    let seeds = vec![Node::new("ic0.app").expect("domain is invalid")];
    let snapshot = LatencyRoutingSnapshot::new();
    let route_provider = DynamicRouteProviderBuilder::new(snapshot, seeds, client.clone()).build().await;
    let transport = ReqwestTransport::create_with_client_route(Arc::new(route_provider), client.clone()).unwrap();

    // Initialize the agent with a custom dynamic transport layer
    let agent = Agent::builder()
        .with_transport(transport)
        .build()
        .expect("failed to create an agent");
    //
    // ... use agent
    //

}
```

# How Has This Been Tested?

1. Added a [mainnet test](https://github.com/dfinity/agent-rs/pull/568/files#diff-c3cffa98e93a87951c04778073900f72c8e3b8225b33b30a9ed2cc88545bb3c8R344)
2. Created a [draft PR](https://github.com/dfinity/ic/pull/516) and [tested](https://dash.zh1-idx1.dfinity.network/invocation/3d9862ab-04c2-431e-b189-3dd506a2bc4f?target=%2F%2Frs%2Ftests%2Fboundary_nodes%3Aapi_boundary_node_decentralization_test&targetStatus=5) on the [api_boundary_node_decentralization_test](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/tests/boundary_nodes/BUILD.bazel?L155), which is a `system-test` in the `ic-repo`.
**NOTE**: @adamspofford-dfinity and @lwshang the suggestion to fully [remove](https://github.com/dfinity/agent-rs/pull/568/commits/ea4c8e5d9e908d45147689e90df2604d30742c61) the `fetch_root_key()` was probably wrong. As `system-test` fails without this call. Could You commit a conditional `fetch_root_key()` invocation, which would work for both `mainnet` and `testnets`.

# Checklist:

### DIFF
--- a/ic-agent/src/agent/http_transport/dynamic_routing/dynamic_route_provider.rs
+++ b/ic-agent/src/agent/http_transport/dynamic_routing/dynamic_route_provider.rs
@@ -263,7 +263,7 @@ mod tests {
         // Only a single node exists, which is initially healthy.
         tokio::time::sleep(snapshot_update_duration).await;
         let routed_domains = route_n_times(6, Arc::clone(&route_provider));
-        assert_routed_domains(routed_domains, vec![node_1.domain.clone()], 6);
+        assert_routed_domains(routed_domains, vec![node_1.domain()], 6);
 
         // Test 2: multiple route() calls return 3 different domains with equal fairness (repetition).
         // Two healthy nodes are added to the topology.
@@ -275,11 +275,7 @@ mod tests {
         let routed_domains = route_n_times(6, Arc::clone(&route_provider));
         assert_routed_domains(
             routed_domains,
-            vec![
-                node_1.clone().domain,
-                node_2.clone().domain,
-                node_3.clone().domain,
-            ],
+            vec![node_1.domain(), node_2.domain(), node_3.domain()],
             2,
         );
 
@@ -288,11 +284,7 @@ mod tests {
         checker.overwrite_healthy_nodes(vec![node_1.clone(), node_3.clone()]);
         tokio::time::sleep(snapshot_update_duration).await;
         let routed_domains = route_n_times(6, Arc::clone(&route_provider));
-        assert_routed_domains(
-            routed_domains,
-            vec![node_1.clone().domain, node_3.clone().domain],
-            3,
-        );
+        assert_routed_domains(routed_domains, vec![node_1.domain(), node_3.domain()], 3);
 
         // Test 4: multiple route() calls return 3 different domains with equal fairness (repetition).
         // Unhealthy node is set back to healthy.
@@ -301,11 +293,7 @@ mod tests {
         let routed_domains = route_n_times(6, Arc::clone(&route_provider));
         assert_routed_domains(
             routed_domains,
-            vec![
-                node_1.clone().domain,
-                node_2.clone().domain,
-                node_3.clone().domain,
-            ],
+            vec![node_1.domain(), node_2.domain(), node_3.domain()],
             2,
         );
 
@@ -323,11 +311,7 @@ mod tests {
         let routed_domains = route_n_times(6, Arc::clone(&route_provider));
         assert_routed_domains(
             routed_domains,
-            vec![
-                node_2.clone().domain,
-                node_3.clone().domain,
-                node_4.clone().domain,
-            ],
+            vec![node_2.domain(), node_3.domain(), node_4.domain()],
             2,
         );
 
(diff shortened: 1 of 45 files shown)


ic-agent/src/agent/http_transport/dynamic_routing/node.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl From<&ApiBoundaryNode> for Node {
+    fn from(api_bn: &ApiBoundaryNode) -> Self {
+        Node::new(api_bn.domain.as_str()).unwrap()
+    }
+}
+
+pub fn is_valid_domain<S: AsRef<str>>(domain: S) -> bool {
+    // TODO
```
r-birkner, line 41: I would just use the same as here: https://github.com/dfinity/ic/blob/master/rs/registry/canister/src/mutations/common.rs#L131-L155

ic-agent/src/agent/http_transport/dynamic_routing/dynamic_route_provider.rs
```diff
@@ -0,0 +1,596 @@
...
+        })?;
+        Ok(node.to_routing_url())
+    }
+}
+
+impl<S> DynamicRouteProvider<S>
+where
+    S: RoutingSnapshot + 'static,
+{
+    pub fn new(snapshot: S, seeds: Vec<Node>, http_client: Client) -> Self {
```
r-birkner, line 78: This means that to create an agent with the discovery library, you need to do something like:
```
let snapshot = LatencyBasedSnapshot::new();
let http_client = Client::builder().build().unwrap();
let route_provider = HealthCheckRouteProvider::new(
            snapshot,
            vec![Node::new(domain_seed)],
            http_client
);
let transport = ReqwestTransport::create_with_client_route(
        Arc::clone(&route_provider) as Arc<dyn RouteProvider>,
        http_client,
    )
    .expect("failed to create transport");
    
let agent = Agent::builder()
        .with_transport(transport)
        .build()?;
```

Can't we do something like and just provide defaults:
```
let agent = Agent::builder()
        .with_latency_based_routing()
        .build()?;
```
or
```
let agent = Agent::builder()
        .with_health_based_routing()
        .build()?;
```

r-birkner: If someone wants not the default, they can then go through the process above and create their own snapshot/route provider, etc.

ic-agent/src/agent/http_transport/dynamic_routing/node.rs
nikolay-komarevskiy: i think reusing the `Url` library some more concise:
```
pub fn is_valid_domain<S: AsRef<str>>(domain: S) -> bool {
    // Prepend scheme to make it a valid URL
    let url_string = format!("http://{}", domain.as_ref());
    Url::parse(&url_string).is_ok()
}
```

ic-agent/src/agent/http_transport/dynamic_routing/dynamic_route_provider.rs
nikolay-komarevskiy: Eventually this is the direction to go, yes! But i don't want to change the `agent` during the library migration phase., better in the follow up.

adamspofford-dfinity (approved):

PR #569, opened 2024-07-14
feat: add support for fetching root keys automatically

rupansh:
# Description

This change introduces a way to fetch root keys automatically on demand. This makes the construction of the agent a bit more ergonomic for local nodes.
Minor breakage to the API had to be introduced unfortunately. I also recommend deprecating the `fetch_root_keys` method.

# How Has This Been Tested?

I ran `cargo test`. Further, this version will also be used by the Yral fronend for testing our changes locally (https://github.com/go-bazzinga/hot-or-not-web-leptos-ssr/blob/rupansh/testcontainers/src/state/canisters.rs#L30)

# Checklist:

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -50,7 +50,10 @@ use std::{
     fmt,
     future::{Future, IntoFuture},
     pin::Pin,
-    sync::{Arc, Mutex, RwLock},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc, Mutex, RwLock,
+    },
     task::{Context, Poll},
     time::Duration,
 };
@@ -259,6 +262,7 @@ pub struct Agent {
     subnet_key_cache: Arc<Mutex<SubnetCache>>,
     concurrent_requests_semaphore: Arc<Semaphore>,
     verify_query_signatures: bool,
+    fetch_root_key_fuse: Arc<AtomicBool>,
     auto_fetch_root_key: bool,
 }
 
@@ -290,6 +294,7 @@ impl Agent {
             subnet_key_cache: Arc::new(Mutex::new(SubnetCache::new())),
             verify_query_signatures: config.verify_query_signatures,
             concurrent_requests_semaphore: Arc::new(Semaphore::new(config.max_concurrent_requests)),
+            fetch_root_key_fuse: Arc::new(AtomicBool::new(false)),
             auto_fetch_root_key: config.auto_fetch_root_key,
         })
     }
@@ -328,7 +333,7 @@ impl Agent {
     /// *Only use this when you are  _not_ talking to the main Internet Computer, otherwise
     /// you are prone to man-in-the-middle attacks! Do not call this function by default.*
     pub async fn fetch_root_key(&self) -> Result<(), AgentError> {
-        if self.read_root_key()[..] != IC_ROOT_KEY[..] {
+        if self.fetch_root_key_fuse.load(Ordering::Acquire) {
             // already fetched the root key
             return Ok(());
         }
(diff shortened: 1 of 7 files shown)


sa-github-api:
Dear @rupansh,

In order to potentially merge your code in this open-source repository and therefore proceed with your contribution, we need to have your approval on DFINITY's [CLA](https://github.com/dfinity/cla/blob/main/CLA.md).

If you decide to agree with it, please visit this [issue](https://github.com/dfinity/cla/issues/488) and read the instructions there. Once you have signed it, re-trigger the workflow on this PR to see if your code can be merged.

— The DFINITY Foundation

adamspofford-dfinity:
Why is this needed vs just calling fetch_root_key immediately after constructing the agent?

rupansh:
> Why is this needed vs just calling fetch_root_key immediately after constructing the agent?

This proposes a more ergonomic API,
fetch_root_key is a part of construction of the agent,
currently this becomes a 2 part process
1. You first configure the agent through the builder
2. You have to further configure the agent through a second call to fetch_root_key.

Other than the ergonomic benefits, the construction of the agent becomes synchronous instead of requiring async.
This is useful if you need a global static agent, or you are in an environment where initializing the agent asynchronously might not be trivial (we encountered this while using the agent in leptos for example)

Imagine a global static agent,
before every call to the replica, one would have to ensure fetch_root_key is called, pervasing the call throughout the code base.

ic-agent/src/agent/mod.rs
```diff
@@ -890,16 +914,17 @@ impl Agent {
         &self,
         delegation: &Option<Delegation>,
         effective_canister_id: Principal,
+        root_key: Vec<u8>,
     ) -> Result<Vec<u8>, AgentError> {
         match delegation {
-            None => Ok(self.read_root_key()),
+            None => Ok(root_key),
```
ericswanson-dfinity, line 920: What is the reason for passing `root_key` as a parameter here, rather than getting it from `self`?

rupansh-gob: verify_cert & check_delegation would become async and you have a recursive future. I wasn't able to unroll it into an iterator unfortunately, so I went with this

rupansh: verify_cert & check_delegation would become async and you have a recursive future. I wasn't able to unroll it into an iterator unfortunately, so I went with this

adamspofford-dfinity:
> Other than the ergonomic benefits, the construction of the agent becomes synchronous instead of requiring async.
This is useful if you need a global static agent

Only for lazy-init statics (e.g. `LazyLock`), yes? Init-once statics (e.g. `OnceLock`) have no trouble being async. And for an agent which may or may not be pointed at the local replica, doesn't this already lend itself to init-once style due to the replica being used being a runtime parameter?

> we encountered this while using the agent in leptos for example

Do you have an example of this?

rupansh:
> > Other than the ergonomic benefits, the construction of the agent becomes synchronous instead of requiring async.
> > This is useful if you need a global static agent
> 
> Only for lazy-init statics (e.g. `LazyLock`), yes? Init-once statics (e.g. `OnceLock`) have no trouble being async. And for an agent which may or may not be pointed at the local replica, doesn't this already lend itself to init-once style due to the replica being used being a runtime parameter?

Indeed

> 
> > we encountered this while using the agent in leptos for example
> 
> Do you have an example of this?

Sure, most of the code is written in a dual-environment style in leptos(i.e targeting both browser and server):
We are providing a global agent here: https://github.com/go-bazzinga/hot-or-not-web-leptos-ssr/blob/rupansh/testcontainers/src/app.rs#L65 

A way to force async initialization would be by using `create_blocking_resource`, but these run server side and need the result to be `Serializable` (which agent is not)
Technically I *can* send the root key server side as a workaround but these are further reduced ergonomics, and introduces latency to the initial load of the website

I understand that this might be too specific to our use case which is why my main emphasis is on the ergonomics rather than better global statics

PR #572, opened 2024-07-17
feat: Add experimental feature for synchronous update calls

DSharifi:
# Description

This PR adds an experimental feature flag to support synchronous update calls, which are routed to `/api/v3/.../call`.  The feature is opt in, and requires the user to also enable a flag in the builder of `ReqwestTransport` or `HyperTransport`.

Changes:
- Route provider no longer provides the `/api/v2` prefix for routes.
- Added the feature flag `experimental_sync_call`, which enables a toggle to `ReqwestTransport` and `HyperTransport` to route ingress messages to the new `v3` call endpoint.
- Agent no longer returns a `RequestId` for update calls, but an enum which either contains the response of the call or the underlying `RequestId` which can be used to poll the request status.

# How Has This Been Tested?

Integration tests have been updated to expect requests are sent to `v3/.../call`.

# Checklist:

### DIFF
--- a/.github/workflows/ic-ref.yml
+++ b/.github/workflows/ic-ref.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install dfx
         uses: dfinity/setup-dfx@main
         with:
-          dfx-version: "latest"
+          dfx-version: "0.22.0-beta.0"
 
       - name: Cargo cache
         uses: actions/cache@v4
--- a/ic-agent/src/agent/http_transport/reqwest_transport.rs
+++ b/ic-agent/src/agent/http_transport/reqwest_transport.rs
@@ -112,7 +112,6 @@ impl ReqwestTransport {
     ) -> Result<(StatusCode, HeaderMap, Vec<u8>), AgentError> {
         let create_request_with_generated_url = || -> Result<Request, AgentError> {
             let url = self.route_provider.route()?.join(endpoint)?;
-            println!("url = {}", url);
             let mut http_request = Request::new(method.clone(), url);
             http_request
                 .headers_mut()
--- a/ic-agent/src/agent/http_transport/reqwest_transport.rs
+++ b/ic-agent/src/agent/http_transport/reqwest_transport.rs
@@ -1,10 +1,9 @@
 //! A [`Transport`] that connects using a [`reqwest`] client.
 #![cfg(feature = "reqwest")]
 
-use std::{sync::Arc, time::Duration};
-
-use ic_transport_types::RejectResponse;
+use ic_transport_types::{RejectResponse, TransportCallResponse};
 pub use reqwest;
+use std::{sync::Arc, time::Duration};
 
 use futures_util::StreamExt;
 use reqwest::{
@@ -19,7 +18,7 @@ use crate::{
         AgentFuture, Transport,
     },
     export::Principal,
-    AgentError, RequestId,
+    AgentError,
 };
 
 /// A [`Transport`] using [`reqwest`] to make HTTP calls to the Internet Computer.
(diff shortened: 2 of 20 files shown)


.github/workflows/ic-ref.yml
DSharifi, line 1: CI would fail, since it would pull `0.21` which uses a replica version that hasn't enabled the new endpoint for synchronous ingress messages.

ic-transport-types/src/lib.rs
```diff
@@ -117,6 +118,36 @@ pub struct ReadStateResponse {
...
+    NonReplicatedRejection(RejectResponse),
+
+    /// The replica timed out the sync request. The status of the request must be polled.
+    Accepted,
+}
+
+/// The response from a request to the `call` endpoint.
+#[derive(Debug, PartialEq, Eq)]
+pub enum CallResponse<Out> {
+    /// The call complted, and the response is available.
```
adamspofford-dfinity, line 144:
```suggestion
    /// The call completed, and the response is available.
```

ic-transport-types/src/lib.rs
```diff
@@ -117,6 +118,36 @@ pub struct ReadStateResponse {
...
+    Response(Out),
+    /// The replica timed out the update call, and the request id should be used to poll for the response
+    /// using the `request_status` request type.
```
adamspofford-dfinity, line 147: Even though it is in ic-transport-types, this should reference an ic-agent function rather than a HTTP endpoint.

ic-agent/src/agent/mod.rs
```diff
@@ -583,16 +564,15 @@ impl Agent {
         })
     }
 
-    /// The simplest way to do an update call; sends a byte array and will return a RequestId.
-    /// The RequestId should then be used for request_status (most likely in a loop).
+    /// The simplest way to do an update call; Sends a byte array and will return a response, [`CallResponse`], from the replica.
```
adamspofford-dfinity, line 567:
```suggestion
    /// The simplest way to do an update call; sends a byte array and will return a response, [`CallResponse`], from the replica.
```

ic-transport-types/src/lib.rs
```diff
@@ -117,6 +118,36 @@ pub struct ReadStateResponse {
     pub certificate: Vec<u8>,
 }
 
+/// The response from a request to the `call` endpoint.
+// TODO: Inline this enum. No need for it to be public.
```
adamspofford-dfinity, line 122: Should this be resolved before release? Arbitrary data structures, even public ones, do not have to go in ic-transport-types, but anything that is part of and deserialized from the replica API can and should.

ic-agent/src/agent/http_transport/reqwest_transport.rs
```diff
@@ -88,6 +89,15 @@ impl ReqwestTransport {
         }
     }
 
+    /// Use call v3 endpoint for update calls.
```
adamspofford-dfinity, line 92: Please explain a little more what the caller is signing up for. (And same for hyper)

ic-transport-types/src/lib.rs
DSharifi: Resolved the stale comment. The enum is used to parse the replica response when it responds `200 - Ok`, however the Accepted variant is manually created if the replica responds with `202 Accepted`.

ic-transport-types/src/lib.rs
```diff
@@ -117,6 +118,36 @@ pub struct ReadStateResponse {
...
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(tag = "status", rename_all = "snake_case")]
+pub enum TransportCallResponse {
+    /// A certified response.
+    Replied {
+        /// The CBOR serialized certificate for the call response.
+        #[serde_as(as = "Bytes")]
```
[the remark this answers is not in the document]
DSharifi: Changed it to serde_bytes

adamspofford-dfinity (approved):

PR #502, opened 2024-01-05
feat!: (ic-utils) support canister upgrade option wasm_memory_persistence

luc-blaeser:
See Motoko's PR on enhanced orthogonal persistence (EOP) more information: https://github.com/dfinity/motoko/pull/4488.

Expose the IC canister upgrade option `wasm_memory_persistence`, such that `dfx` can pass this flag when upgrading Motoko canisters with EOP.

PR #573, opened 2024-07-22
chore: Bump MSRV to 1.75.0

adamspofford-dfinity:
`rust-analyzer` no longer supports versions below 1.75, so it is no longer reasonable to develop for 1.70.


lwshang:
MSRV has been updated in #502

adamspofford-dfinity:
whoops, missed that

PR #598, opened 2024-09-23
feat: Add HTTP middleware for Transport replacement

adamspofford-dfinity:
ic-gateway was broken by #586, this provides a middle ground for them.

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -182,8 +182,7 @@ impl Agent {
             ingress_expiry: config.ingress_expiry.unwrap_or(DEFAULT_INGRESS_EXPIRY),
             root_key: Arc::new(RwLock::new(IC_ROOT_KEY.to_vec())),
             client: config.http_service.unwrap_or_else(|| {
-                Arc::new(DefaultRetryLogic {
-                    _max_tcp_error_retries: config.max_tcp_error_retries,
+                Arc::new(Retry429Logic {
                     client: config.client.unwrap_or_else(|| {
                         #[cfg(not(target_family = "wasm"))]
                         {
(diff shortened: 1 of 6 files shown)


ic-agent/src/agent/mod.rs
```diff
@@ -1868,6 +1841,90 @@ impl<'agent> IntoFuture for UpdateBuilder<'agent> {
     }
 }
 
+/// HTTP client middleware. Implemented automatically for `reqwest`-compatible by-ref `tower::Service`, such as `reqwest_middleware`.
+pub trait HttpService: Send + Sync {
```
blind-oracle, line 1845: Is there some reason we don't use `async_trait` here to make it simpler? We can work with it, but just wondering...

ic-agent/src/agent/mod.rs
```diff
@@ -1868,6 +1841,90 @@ impl<'agent> IntoFuture for UpdateBuilder<'agent> {
...
+impl HttpService for Retry429Logic {
+    fn call<'a>(
+        &'a self,
+        req: &'a (dyn Fn() -> Result<Request, AgentError> + Send + Sync),
+        _max_retries: usize,
+    ) -> AgentFuture<'a, Response> {
+        Box::pin(async move {
+            loop {
+                #[cfg(not(target_family = "wasm"))]
+                let resp = self.client.call(req, _max_retries).await?;
```
nikolay-komarevskiy, line 1913: Maybe add `max_retries` for 429 instead of retrying indefinitely?

ic-agent/src/agent/mod.rs
adamspofford-dfinity: No reason, I suppose.

ic-agent/src/agent/mod.rs
adamspofford-dfinity: The reason for retrying 429s indefinitely is because you are not served them indefinitely - even though it means the semaphore isn't doing its job, your message backlog will eventually be cleared, or if it grows too much, then you'll start getting message expiry timeouts. Is this bad logic?

lwshang (approved):

PR #599, opened 2024-09-23
feat: unstable dynamic routing feature

adamspofford-dfinity:
The final design of dynamic routing in the agent is not complete and other teams want to be already using it, so this merges the original design with the minimum number of changes to support WASM, and exports it under the feature `_internal_dynamic-routing`. Work on the final design will continue in the background.


lwshang (approved):

ic-agent/src/agent/route_provider/dynamic_routing/dynamic_route_provider.rs
```diff
@@ -7,19 +7,15 @@ use std::{
 
 use arc_swap::ArcSwap;
 use candid::Principal;
+use futures_util::FutureExt;
 use reqwest::Client;
+use stop_token::StopSource;
 use thiserror::Error;
-use tokio::{
```
nikolay-komarevskiy: @adamspofford-dfinity even though `futures_util` can run with `tokio` runtime, i'm not sure the performance will be as good as with pure `tokio`.  As `tokio` specific utilities like tasks are optimized for `tokio`, so i'm a bit concerned with the decision to remove it.

nikolay-komarevskiy: i get the rationale to make the code work for `wasm` builds too, but don't like that standard `build` will potentially loose in performance

PR #606, opened 2024-10-04
feat: Use `ed25519-consensus` instead of `ring`

adamspofford-dfinity:
ring's API is very unfriendly and prevents e.g. cloning or serializing. ic-agent already uses ed25519-consensus, so this essentially removes a redundant dependency. The existing function is placed under a feature flag.


ic-transport-types/src/lib.rs
```diff
@@ -17,7 +17,7 @@ mod request_id;
 pub mod signed;
 
 /// The authentication envelope, containing the contents and their signature. This struct can be passed to `Agent`'s
-/// `*_signed` methods via [`to_bytes`](Envelope::to_bytes).
+/// `*_signed` methods via [`to_bytes`](Envelope::encode_bytes).
```
lwshang, line 20:
```suggestion
/// `*_signed` methods via [`encode_bytes`](Envelope::encode_bytes).
```

lwshang (approved):

PR #607, opened 2024-10-11
chore: run integration tests against PocketIC

mraszyk:
This PR runs integration tests against PocketIC in addition to the local replica.


ref-tests/src/utils.rs
```diff
@@ -11,8 +11,18 @@ const HSM_SLOT_INDEX: &str = "HSM_SLOT_INDEX";
 const HSM_KEY_ID: &str = "HSM_KEY_ID";
 const HSM_PIN: &str = "HSM_PIN";
 
-pub fn get_effective_canister_id() -> Principal {
-    Principal::from_text("rwlgt-iiaaa-aaaaa-aaaaa-cai").unwrap()
+pub async fn get_effective_canister_id(agent: &Agent) -> Principal {
+    match agent.fetch_pocketic_topology().await {
+        Ok(topology) => {
+            let app_subnet = topology.get_app_subnets()[0];
```
chenyan-dfinity, line 17: If `networks.json` specifies the system subnet type, would this work?

mraszyk: > If networks.json specifies the system subnet type, would this work?

No, but this function is supposed to only be used internally in ic-agent tests, not externally by ic-agent users.

adamspofford-dfinity (approved):

PR #632, opened 2025-01-29
feat!: Lowercase `WasmMemoryPersistence`

adamspofford-dfinity:
Match the new replica's management canister API.


ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -555,6 +567,22 @@ pub enum InstallMode {
...
+                opts.skip_pre_upgrade.get_or_insert(false);
+                opts.wasm_memory_persistence
+                    .get_or_insert(WasmMemoryPersistence::Replace);
+                Self::Upgrade(Some(opts))
+            }
+            x => *x,
+        }
+    }
+}
+
```
lwshang, line 585: Could we keep the fields as-is without the `canonicalize()` method? I believe optional fields should retain their `None` values rather than being transformed.

lwshang (approved):

PR #633, opened 2025-02-02
fix: AgentBuilder::with_background_dynamic_routing

zensh:
# Description

Remove redundant `async` from the `with_background_dynamic_routing` function.


ericswanson-dfinity:
Change merged in https://github.com/dfinity/agent-rs/pull/634

PR #653, opened 2025-07-16
feat: add environment variables to canister settings

rikonor:
See https://dfinity.atlassian.net/browse/SDK-2151.

This change is required to add support for canister environment variables in the 
new `icp-cli` tool.

### DIFF
--- a/.github/workflows/ic-ref.yml
+++ b/.github/workflows/ic-ref.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Install dfx
         uses: dfinity/setup-dfx@main
         with:
-          dfx-version: "0.24.3"
+          dfx-version: "0.28.0"
 
       - name: Cargo cache
         uses: actions/cache@v4
--- a/ic-utils/src/interfaces/management_canister.rs
+++ b/ic-utils/src/interfaces/management_canister.rs
@@ -179,8 +179,8 @@ pub enum LogVisibility {
 /// A generic environment variable struct defining a key-value pair, meant to be used for defining a canister environment
 #[derive(Debug, Clone, CandidType, Deserialize, PartialEq)]
 pub struct EnvironmentVariable {
-    /// Variable key/name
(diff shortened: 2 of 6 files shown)


lwshang:
The field name must be consistent with the spec:
```
type canister_settings = record {
    ...
    environment_variables : opt vec environment_variable;
}
```

So we should use `environment_variables` instead of `environment`.

For consistency, I also suggest these naming changes, though they don't affect correctness:
- Rename the key-value pair type : `Variable` -> `EnvironmentVariable`;
- Rename the builder methods: `with_environment_variables`, `with_optional_environment_variables`.

During my Rust CDK work, I found Pocket-IC doesn't support this feature yet, so we can't properly test it right now. The incorrect field name was actually skipped during Candid deserialization since it's an optional field, meaning our existing tests (including those in the PR) wouldn't catch this error.

rikonor:
> During my Rust CDK work, I found Pocket-IC doesn't support this feature yet, so we can't properly test it right now.

I spoke with @mraszyk yesterday about this, and he referred me to this [PR](https://github.com/dfinity/ic/pull/5951) which should bring support to pocket-ic. I will try to test with it shortly and report back.

rikonor:
> The field name must be...

Thanks for the feedback @lwshang, I'll make the changes and report back 👍

mraszyk:
The PR is now merged: please use master commits/artifacts for PocketIC.

lwshang:
> would you be able to provide some guidance on how to go about adding an integration test to this repo?

Sure. The `ref-tests` crate contains the integration tests of the repo. The `ic-utils` related tests are in this [file](https://github.com/dfinity/agent-rs/blob/main/ref-tests/tests/ic-ref.rs).

For guidance on how to run the tests locally, you can refer to this [CI workflow](https://github.com/dfinity/agent-rs/blob/main/.github/workflows/ic-ref.yml).

PR #663, opened 2025-10-06
feat: initial draft of ic-utils-bindgen

adamspofford-dfinity:


Cargo.toml
```diff
@@ -31,7 +32,7 @@ ic-transport-types = { path = "ic-transport-types", version = "0.44.1" }
 
 ic-certification = "3"
 candid = "0.10.10"
-candid_parser = "0.1.4"
+candid_parser = "0.2.0"
```
ilbertt, line 35: Latest is `0.2.2`.

```suggestion
candid_parser = "0.2.2"
```

ilbertt (changes requested):
Just noticed this, I haven't looked into the implementation.

Cargo.toml
adamspofford-dfinity: I'll change it because you asked, but for future reference this doesn't matter, it'd be impossible for anyone to end up with 0.2.0 because the latest version is always selected.

PR #667, opened 2025-10-15
docs: Fix docs build

adamspofford-dfinity:
`feature(doc_auto_cfg)` has been merged into `feature(doc_cfg)`. Also removes manual setting of `cfg(docsrs)` because docs.rs includes it automatically.


lwshang (approved):
Great work on getting the docs.rs build check integrated into CI. I know that turned out to be less trivial than expected, so thanks for sticking with it!

PR #731, opened 2026-05-14
chore: release 0.47.3 — subnet-scoped create_canister + EffectiveId routing

lwshang:
## Summary

- Introduces a public `EffectiveId { Canister(Principal) | Subnet(Principal) }` enum in `ic-agent::agent`; bare `Principal` still works via `From<Principal>`, so existing call sites are source-compatible.
- `Agent::update_signed`, `request_status_signed`, and `wait_signed` now take `impl Into<EffectiveId>` and dispatch internally to the canister-scoped or subnet-scoped HTTP endpoints (`/api/v4/subnet/<id>/call`, `/api/v3/subnet/<id>/read_state`). No duplicated `_subnet` call stack.
- Adds `CreateCanisterBuilder::with_effective_subnet_id` so subnet administrators can route `create_canister` to a specific subnet. The builder signs the management-canister update via `UpdateBuilder::sign()` and submits it through the subnet-scoped path; the canister-id path is unchanged.
- Renames the `sign_request_status` parameter `effective_canister_id` → `effective_id` and documents that `SignedRequestStatus::effective_canister_id` may carry a subnet id when the request was originally subnet-scoped.
- Bumps all workspace crates to 0.47.3.

## Test plan

- [x] Existing ic-ref tests pass.
- [x] New ref-test boots PocketIC with a `cloud_engine` subnet whose `subnet_admins` contains the agent identity and asserts that `create_canister().with_effective_subnet_id(...)` lands the canister on that subnet.

### DIFF
--- a/ic-agent/src/agent/mod.rs
+++ b/ic-agent/src/agent/mod.rs
@@ -381,6 +381,24 @@ impl Agent {
         serde_cbor::from_slice(&response_body).map_err(AgentError::InvalidCborData)
     }
 
+    async fn call_subnet_endpoint(
+        &self,
+        subnet_id: Principal,
+        serialized_bytes: Vec<u8>,
+    ) -> Result<TransportCallResponse, AgentError> {
+        let _permit = self.concurrent_requests_semaphore.acquire().await;
+        let endpoint = format!("api/v4/subnet/{}/call", subnet_id.to_text());
+        let (status_code, response_body) = self
+            .execute(Method::POST, &endpoint, Some(serialized_bytes))
+            .await?;
+
+        if status_code == StatusCode::ACCEPTED {
+            return Ok(TransportCallResponse::Accepted);
+        }
+
+        serde_cbor::from_slice(&response_body).map_err(AgentError::InvalidCborData)
+    }
+
     /// The simplest way to do a query call; sends a byte array and will return a byte vector.
     /// The encoding is left as an exercise to the user.
     #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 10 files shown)


lwshang:
The Ubuntu CI failures in `identity::delegated_canister_sig_identity` and `identity::sender_info_via_canister_sig` are caused by a bug in pocket-ic (introduced in the same IC rev pinned by this PR).

**Root cause:** `get_delegation_for_subnet` in `rs/state_machine_tests/src/lib.rs` does not include the `/subnet/<id>/type` path when building delegation certs, so all pocket-ic-issued delegations have `subnet_type = None`. A recent change to `rs/certification/src/lib.rs` now rejects delegations where `subnet_type` is `None`, causing canister-signature validation to fail with _"the source subnet is not trusted for delegations"_.

**Fix:** https://github.com/dfinity/ic/pull/10226 — once merged, this PR can be rebased on a newer IC rev that includes the fix.

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -360,47 +462,102 @@ impl<'agent, 'canister: 'agent> CreateCanisterBuilder<'agent, 'canister> {
...
+async fn subnet_call_and_wait(
+    canister: &Canister<'_>,
+    method: &str,
+    arg_bytes: Vec<u8>,
+    subnet_id: Principal,
+) -> Result<(Principal,), AgentError> {
+    let agent = canister.agent;
+    let signed = agent
+        .update(&Principal::management_canister(), method)
+        .with_arg(arg_bytes)
```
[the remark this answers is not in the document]
lwshang: Thanks for the review. On closer inspection,  in ic-agent takes `A: Into<Vec<u8>>` and simply stores the bytes as-is — no Candid encoding is applied at that layer. The `arg_bytes` here are passed through without any double-encoding, so the original code is correct.

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -285,7 +325,74 @@ impl<'agent, 'canister: 'agent> CreateCanisterBuilder<'agent, 'canister> {
...
     pub fn build(self) -> Result<impl 'agent + AsyncCall<Value = (Principal,)>, AgentError> {
+        let prepared = self.prepare()?;
+        let effective_canister_id = match prepared.effective_id {
+            EffectiveId::Canister(p) => p,
+            EffectiveId::Subnet(_) => {
+                return Err(AgentError::MessageError(
+                    "CreateCanisterBuilder::build() does not support subnet routing; \
+                     use call()/call_and_wait() instead."
+                        .into(),
+                ));
```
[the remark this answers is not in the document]
lwshang: Good catch. Fixed in eee7917 — the `AsyncCall` impl now delegates to the inherent `call`/`call_and_wait` methods (which handle both canister and subnet routing) instead of going through `build()`. `IntoFuture` is updated similarly to `Box::pin(self.call_and_wait())`.

lwshang:
Bumped the pocket-ic rev to pick up the state_machine_tests delegation fix from [dfinity/ic#10226](https://github.com/dfinity/ic/pull/10226). The two previously failing canister-signature tests now pass locally.

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -95,20 +108,47 @@ impl<'agent, 'canister: 'agent> CreateCanisterBuilder<'agent, 'canister> {
         Self {
             is_provisional_create: true,
             specified_id: Some(specified_id),
-            effective_canister_id: specified_id,
+            effective_id: EffectiveId::Canister(specified_id),
```
adamspofford-dfinity, line 108: This silently overwrites a potential previous effective subnet ID; this should be documented or panicked upon

ic-agent/src/agent/mod.rs
```diff
@@ -773,6 +858,61 @@ impl Agent {
         }
     }
 
+    /// Sibling of [`wait_signed`](Self::wait_signed) that polls the subnet-scoped
+    /// read_state endpoint (`/api/v3/subnet/<subnet_id>/read_state`).
+    pub async fn wait_signed_subnet(
```
adamspofford-dfinity: Is it worth lifting EffectiveId to ic-agent, so we don't have to duplicate the whole call-stack? That's what I've settled on in icp-js-core.

adamspofford-dfinity (commented):
This seems to be missing the subnet query call implementation.

ic-utils/src/interfaces/management_canister/builders.rs
lwshang: Good point. Addressed in 8970f76 — the doc comments on `as_provisional_create_with_specified_id`, `with_effective_canister_id`, and `with_effective_subnet_id` now all explicitly note that the three share the same field and only the last call takes effect; earlier ones are silently shadowed. I went with documenting rather than panicking since shadowing is a normal builder pattern and a panic would be surprising for users who deliberately reconfigure mid-chain.

ic-agent/src/agent/mod.rs
lwshang: Already done — `EffectiveId` is defined in `ic-agent` ([ic-agent/src/agent/mod.rs:93](https://github.com/dfinity/agent-rs/blob/lwshang/SDK-2702_create_canister_subnet_id/ic-agent/src/agent/mod.rs#L93)) and re-exported under `ic_agent::agent::EffectiveId`. `ic-utils` just imports it; no duplication.

adamspofford-dfinity (approved):

PR #737, opened 2026-07-13
chore: release 0.49.0 — ic-management-canister-types 0.8.0, query-default reads, list_canisters

lwshang:
Release 0.49.0. Bumps `ic-management-canister-types` 0.7.1 → 0.8.0 and builds on the new types/methods, plus a couple of related `ic-utils` ergonomics. All workspace crates go 0.48.1 → 0.49.0.

## `ic-utils` — management canister

- **`ic-management-canister-types` 0.8.0.** Picks up the `snapshot_visibility` canister setting and the `list_canisters` / `canister_metrics` types.
  - New `with_snapshot_visibility` setter on `CreateCanisterBuilder` and `UpdateSettingsBuilder`.
  - Re-exports `SnapshotVisibility`, `CanisterIdRange`, `ListCanistersResult`, `CyclesConsumed`, `CanisterMetricsArgs`, `CanisterMetricsResult`.
- **`canister_metrics()`** — new `ManagementCanister` read.
- **`list_canisters()`** — subnet-scoped, query-only management method (subnet-administrators only per the interface spec). Routed to the subnet-scoped query endpoint via `Agent::query_signed` with `EffectiveId::Subnet` — reusing the existing agent mechanism rather than adding a `QueryBuilder::with_effective_subnet_id`, keeping update/query API symmetry.
- **Query-default reads.** `canister_status` and `canister_metrics` — the only two management reads the replica accepts as both query and update — now return a `QueryOrUpdateCall` that issues a cheap, non-replicated query by default (`.call().await`), with `.as_update().call().await` to opt into a replicated, certified call. Query responses are replica-signed and verified by the agent unless verification is disabled. `fetch_canister_logs` and `list_canisters` stay query-only per spec.
- **`with_canister_settings(CanisterSettings)`** on `CreateCanisterBuilder` and `UpdateSettingsBuilder`, for callers that already hold a full settings struct (e.g. forwarding one, as icp-cli's `create_mgmt` does). Currently mutually exclusive with the individual `with_*` setters (building errors if combined); can be loosened to a merge model later without breaking callers.

## Breaking changes

- Re-exported `CanisterSettings` / `DefiniteCanisterSettings` gained a `snapshot_visibility` field; neither is `#[non_exhaustive]`, so struct-literal construction must add it.
- `canister_status` now returns `QueryOrUpdateCall` (query by default) instead of `impl AsyncCall` (replicated update). Migrate `.call_and_wait().await` → `.call().await`, or `.as_update().call().await` to keep the old behavior.
- `MgmtMethod` gained `CanisterMetrics` and `ListCanisters` variants; it is not `#[non_exhaustive]`, so exhaustive matches must add arms.

See [CHANGELOG.md](CHANGELOG.md) for the full list.

## Testing / infra

- Bumped the ref-tests `pocket-ic` git rev to `release-2026-07-09_04-35-base` (pocket-ic 13.0.0 → 15.0.0) and the `icx` crates.io pin to 15.0.0 to keep the major in sync. Re-tuned `provisional_create_canister_with_cycles` for pocket-ic 15's stricter memory reserved-cycles floor.
- New (non-ignored) ref-tests for `canister_metrics`, `list_canisters` (subnet-admin harness), and `with_canister_settings` (happy path + mutual-exclusion error). The `canister_metrics` test exercises both the query and `.as_update()` paths.
- Full ref-tests suite green (ic-ref 36, integration 17); workspace `clippy --all-targets` and `fmt --check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -391,58 +427,108 @@ impl<'agent, 'canister: 'agent> CreateCanisterBuilder<'agent, 'canister> {
...
+            || self.environment_variables.is_some()
+            || self.snapshot_visibility.is_some();
+
+        if let Some(settings) = self.settings.take() {
+            if has_individual_settings {
+                return Err(AgentError::MessageError(
+                    "`with_canister_settings` cannot be combined with the individual settings \
+                     setters (`with_controller`, `with_compute_allocation`, etc.)"
+                        .to_string(),
+                ));
```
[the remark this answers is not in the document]
lwshang: Not a defect. Rust's string-continuation escape — a `\` at the end of a line inside a string literal — removes the newline **and all leading whitespace on the next line**, so the indentation is stripped at compile time and never reaches the runtime message. Compiled and verified: the emitted string is `` `with_canister_settings` cannot be combined with the individual settings setters (...)`` — a single space between "settings" and "setters" (the intentional space kept before the `\`), with no embedded indentation. Leaving as-is.

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -1213,83 +1280,154 @@ impl<'agent, 'canister: 'agent> UpdateSettingsBuilder<'agent, 'canister> {
...
```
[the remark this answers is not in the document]
lwshang: Not a defect. Rust's string-continuation escape — a `\` at the end of a line inside a string literal — removes the newline **and all leading whitespace on the next line**, so the indentation is stripped at compile time and never reaches the runtime message. Compiled and verified: the emitted string is `` `with_canister_settings` cannot be combined with the individual settings setters (...)`` — a single space between "settings" and "setters" (the intentional space kept before the `\`), with no embedded indentation. Leaving as-is.

icx/src/main.rs
```diff
@@ -334,6 +335,13 @@ pub fn get_effective_canister_id(
             | MgmtMethod::CanisterInfo => {
                 bail!("Management canister method {method_name} can only be run from canisters");
             }
+            MgmtMethod::ListCanisters => {
+                bail!(
+                    "Management canister method {} is subnet-scoped and cannot be routed by \
+                     effective canister id; it is not supported through icx.",
+                    method_name.as_ref()
+                );
```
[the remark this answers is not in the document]
lwshang: Not a defect. Rust's string-continuation escape (`\` at end of line in a string literal) strips the newline **and the leading whitespace of the next line** at compile time, so no indentation reaches the emitted message. Verified the string is `Management canister method <m> is subnet-scoped and cannot be routed by effective canister id; it is not supported through icx.` — single spaces throughout. Leaving as-is.

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -391,58 +427,108 @@ impl<'agent, 'canister: 'agent> CreateCanisterBuilder<'agent, 'canister> {
         }
     }
 
-    fn prepare(self) -> Result<PreparedCreate<'agent, 'canister>, AgentError> {
-        let controllers = match self.controllers {
+    /// Resolve the builder's settings into a single [`CanisterSettings`].
+    ///
+    /// Returns an error if [`Self::with_canister_settings`] was combined with any
+    /// of the individual settings setters, which are currently mutually exclusive.
+    fn resolve_settings(&mut self) -> Result<CanisterSettings, AgentError> {
```
adamspofford-dfinity: The documentation does not mention that this function clears all the settings from the receiver, so that awaiting it will not update any settings. (Should it? This function doesn't seem like a good idea to me. Consuming the receiver has exactly the right semantics here.)

ic-utils/src/interfaces/management_canister/builders.rs
```diff
@@ -1213,83 +1280,154 @@ impl<'agent, 'canister: 'agent> UpdateSettingsBuilder<'agent, 'canister> {
...
+            settings: Some(settings),
+            ..self
+        }
+    }
+
+    /// Resolve the builder's settings into a single [`CanisterSettings`].
+    ///
+    /// Returns an error if [`Self::with_canister_settings`] was combined with any
+    /// of the individual settings setters, which are currently mutually exclusive.
+    fn resolve_settings(&mut self) -> Result<CanisterSettings, AgentError> {
```
adamspofford-dfinity: Same comment as above

ic-utils/src/interfaces/management_canister/builders.rs
lwshang: Good catch — fixed in 0553f32. Removed the `resolve_settings(&mut self)` helper entirely and inlined the settings resolution into `prepare()` (create) and `build()` (update), so `self` is consumed by value and the fields are moved out via `match` — no more `&mut self` + `.take()` gutting the receiver. The only shared bit left is a tiny `reject_mixed_settings(bool)` helper for the mutual-exclusivity error message (to avoid duplicating the string across both builders). This also matches the pre-existing style in this file, where the per-field conversion was already inlined in both `prepare()` and `build()`.

ic-utils/src/interfaces/management_canister/builders.rs
lwshang: Fixed together with the create-side comment in 0553f32 — the `UpdateSettingsBuilder` resolution is now inlined into `build()` the same way (no `&mut self`/`take`).

adamspofford-dfinity (approved):

